PR #14, opened 2023-04-11
fix: fix issue with blank coordinate columns (cf workflow 2023-03-29)

Pierlou:
Previous behavior: for each file that fits a schema, keep only the columns registered in the schema version and fill the missing columns with blanks (this resulted in blanks in the coordinates column of IRVE files, therefore making the DAG crash in the geo task)

New behavior: for each file that fits a schema, keep only the columns registered in the schema version. Check that all required columns (from the `constraints` set in the schema) are in the file. If so, add potential missing columns with blanks and proceed. Otherwise, discard the file.

### DIFF
--- a/schema/scripts/schemas_consolidation/schemas_consolidation.py
+++ b/schema/scripts/schemas_consolidation/schemas_consolidation.py
@@ -11,6 +11,7 @@ import os
 import yaml
 from datetime import datetime
 import time
+pd.set_option('display.max_columns', None)
 
 
 VALIDATA_BASE_URL = (
@@ -147,6 +148,8 @@ def parse_api_search(url: str, api_url: str, schema_name: str) -> pd.DataFrame:
 # API parsing to get resources infos based on schema metadata, tags and search keywords
 def parse_api(url: str, schema_name: str) -> pd.DataFrame:
     r = requests.get(url)
+    # better error catch in case schema is not responding
+    r.raise_for_status()
     data = r.json()
     nb_pages = int(data["total"] / data["page_size"]) + 1
     arr = []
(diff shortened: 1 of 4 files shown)


geoffreyaldebert (approved):
👏 but i do not understand why this change generate +10k irve. New data since bug ?

AntoineAugusti:
Not super familiar with this DAG/this particular schema but looks sensible to me!

schema/scripts/schemas_consolidation/schemas_consolidation.py
```diff
@@ -690,16 +688,24 @@ def run_schemas_consolidation(
...
+                        version_all_cols_list = [
                             field_dict["name"] for field_dict in version_dict["fields"]
                         ]
+                        version_required_cols_list = [
+                            field_dict["name"] for field_dict in version_dict["fields"]
+                            if (
+                                "constraints" in field_dict.keys()
+                                and "required" in field_dict["constraints"].keys()
+                                and field_dict["constraints"]["required"]
+                            )
```
maudetes, line 707: Could be simplified to this?
```suggestion
                            if field_dict.get("constraints", {}).get("required")
```

maudetes (commented):
Seems fine! :) As @geoffreyaldebert mentions, it doesn't seem to explain why the new lines though

Pierlou:
I added more safeguards, but fun fact 🤡 in this file _https://www.data.gouv.fr/fr/datasets/r/67ed303d-1b3a-49d1-afb4-6c0e4318cc20_ there are some cells with "\n", which `pandas.to_csv` apparently has a hard time with, and that seems to be what broke the DAG in the first place. Ready to push to production if you all agree.
PS: for the new lines added, I don't know where they come from exactly. My best guess is that some files were not taken into consideration previously and they now appear in the consolidated dataset (FYI: it's not duplicated data, and the new lines pass the schema checks)

PR #16, opened 2023-04-27
DAG RNA

Pierlou:
DAG agrégation des données RNA, passage dans csv-detective et publication en tant que ressource communautaire


geoffreyaldebert:
This PR should only touch RNA folder and in it there is removal of DVF data. Please rebase and propose changes for only RNA data 🙏 :)

geoffreyaldebert (approved):

PR #18, opened 2023-05-05
[data_processing] [sirene] feat: prevent new rate limiting from api insee

geoffreyaldebert:
wait 0.5 sec between each call insee


geoffreyaldebert:
FYI @HAEKADI

PR #22, opened 2023-05-22
Feat/add matomo metrics

maudetes:


geoffreyaldebert (approved):
seems good as it is mainly my work

PR #28, opened 2023-06-22
create whole period stats and change DPE download

Pierlou:


Pierlou:
@geoffreyaldebert I think I've updated everything we talked about

geoffreyaldebert (approved):

PR #29, opened 2023-06-26
Use the detected content-type to include a resource or not

AntoineAugusti:
Use the attribute `check:headers:content-type` stored in the `extras` to know if we should include a resource in a consolidation or not.

Previously we only looked at extensions present in the URL of the resource. Some URLs don't end in `.csv` and therefore were excluded.

### DIFF
--- a/schema/scripts/schemas_consolidation/schemas_consolidation.py
+++ b/schema/scripts/schemas_consolidation/schemas_consolidation.py
@@ -115,6 +115,7 @@ def parse_api_search(url: str, api_url: str) -> pd.DataFrame:
                     else:
                         filename = res["url"].split("/")[-1]
                     ext = filename.split(".")[-1]
+                    detected_mime = res.get("extras", {}).get("check:headers:content-type", "").split(";").trim()
                     obj = {}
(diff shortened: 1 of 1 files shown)


schema/scripts/schemas_consolidation/schemas_consolidation.py
```diff
@@ -115,6 +115,7 @@ def parse_api_search(url: str, api_url: str) -> pd.DataFrame:
                     else:
                         filename = res["url"].split("/")[-1]
                     ext = filename.split(".")[-1]
+                    detected_mime = res.get("extras", {}).get("check:headers:content-type", "").split(";").trim()
```
Pierlou, line 118: Not sure what you are expecting from this: `res.get("extras", {}).get("check:headers:content-type", "").split(";")` returns a list, what is `trim`?

AntoineAugusti: My bad https://github.com/etalab/datagouvfr_data_pipelines/pull/29/commits/e58feaa6e9a67553d1c97a47050188c99f7f5f27

PR #54, opened 2023-07-25
DAG geozones

Pierlou:
First part of https://github.com/etalab/data.gouv.fr/issues/1116


utils/datagouv.py
```diff
@@ -104,6 +103,10 @@ def post_resource(
         url = f"{DATAGOUV_URL}/api/1/datasets/{dataset_id}/upload/"
     r = requests.post(url, files=files, headers=headers)
     r.raise_for_status()
+    if not resource_id:
```
maudetes, line 106: Not sure I've understood the logic, but if it works... :shrug:

data_processing/geozones/scripts/script_dl_geozones.sh
```diff
@@ -0,0 +1,10 @@
+DATADIR="$1"
```
maudetes, line 1: Is this file currently used?

maudetes (approved):
I have minor comments, but seems good!:muscle:

data_processing/geozones/task_functions.py
```diff
@@ -0,0 +1,264 @@
...
+            "nomSansArticle": "DROM-COM",
+            "codeArticle": None,
+            "type": "country-subset",
+            "is_deleted": False,
+            "level": "country-subset",
+            "_id": "country-subset:fr:dromcom"
+        },
+        {
+            "uri": None,
+            "nom": "Metropolitan France",
```
maudetes, line 154: I suggest to use the French name:
```suggestion
            "nom": "France métropolitaine",
```

PR #58, opened 2023-08-16
Meta DAG

Pierlou:


meta/config/config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "dag_list": [
```
maudetes, line 2: Do we want to have an "*" option? Or should we keep a declarative list?

maudetes (approved):
Seems nice! Just a suggestion on dag list :)

PR #63, opened 2023-08-24
[Meta DAG] Set time to UTC+2

Pierlou:


meta/task_functions.py
```diff
@@ -69,19 +69,28 @@ def notification_mattermost(ti):
             hours = int(average_duration // 3600)
             minutes = int((average_duration % 3600) // 60)
             start_time = sorted(successes.keys())[-1].split(' ')[1][:-3]
+            # setting time to UTC+2
+            start_hour = int(start_time.split(':')[0])
```
maudetes, line 73: Wouldn't it be easier to convert it to datetimes to manipulate instead?

Pierlou: Idk if it's easier, I guess more elegant indeed

maudetes: less error-prone I would say

maudetes (approved):
Nice! Looks good to me!

PR #65, opened 2023-09-01
[DAG consolidation] hydra extras not used for remote resources

Pierlou:


utils/schema.py
```diff
@@ -177,36 +177,50 @@ def parse_api(url: str, api_url: str, schema_name: str) -> pd.DataFrame:
...
-        last_modification_date = datetime.fromisoformat(extras["analysis:last-modified-at"])
-        last_validation_date = datetime.fromisoformat(extras["validation-report:validation_date"])
-        # progressively switching to timezone-aware dates
-        if not last_validation_date.tzinfo:
-            last_validation_date = local_timezone.localize(last_validation_date)
-        if last_modification_date > last_validation_date:
-            print("recent hydra check: validation")
-            # resource has been changed since last validation: validate again
+    data = requests.get(resource_api_url).json()
+    # if resource is a file on data.gouv.fr (not remote, due to hydra async work)
```
geoffreyaldebert, line 181:
```suggestion
    # if resource is a file on data.gouv.fr (not remote, due to hydra async work)
    # Today (2023-09-04), hydra process a check every week and we want a consolidation every day,
    #  this condition should be removed when hydra and consolidation follows the same schedule (every day).
    # Not for now
```

geoffreyaldebert (approved):

PR #69, opened 2023-09-12
DAG indicateurs d'impact data.gouv

Pierlou:


dgv/impact/task_functions.py
```diff
@@ -0,0 +1,188 @@
...
+    average_quality_score = round(100 * final["quality_score"].mean(), 2)
+
+    # time for legitimate answer
+    print("Calculating average time for legitimate answer")
+    datagouv_team = requests.get(
+        "https://www.data.gouv.fr/api/1/organizations/646b7187b50b2a93b1ae3d45/"
+    ).json()
+    datagouv_team = [m['user']['id'] for m in datagouv_team['members']]
+
+    r = get_all_from_api_query("https://www.data.gouv.fr/api/1/discussions/?sort=-created")
```
maudetes, line 53: You may use the discussion catalog: https://www.data.gouv.fr/fr/datasets/catalogue-des-donnees-de-data-gouv-fr/#/resources/d77705e1-4ecd-461c-8c24-662d47c4c2f9.
You won't have all infos, but since you do an additional call, I think it would be enough to filter on end and start date and messages length?

dgv/impact/task_functions.py
```diff
@@ -0,0 +1,188 @@
...
+                    f"https://www.data.gouv.fr/api/1/datasets/{discussion['subject']['id']}/"
+                ).json()
+                if dataset.get('organization', None):
+                    dataset_supervisors = requests.get(
+                        f"https://www.data.gouv.fr/api/1/organizations/{dataset['organization']['id']}/"
+                    ).json()
+                    dataset_supervisors = [m['user']['id'] for m in dataset_supervisors['members']]
+                else:
+                    dataset_supervisors = [dataset['owner']['id']] if dataset['owner'] else []
+                legit = datagouv_team + dataset_supervisors
```
maudetes, line 83: It seems readable enough :+1:

maudetes (approved):
Not tested, but seems clear! :ok_hand:

dgv/impact/task_functions.py
Pierlou: As discussed, all the infos I need are not in this dataset so I will have to make API calls for each discussion anyway. I can use the dataset to filter on creation dates however, it changes the first part but not the rest of it. Could replace lines 53-65 with:
```
df_discussions = pd.read_csv('https://www.data.gouv.fr/fr/datasets/r/d77705e1-4ecd-461c-8c24-662d47c4c2f9', sep=';')
end_date = datetime.today().strftime("%Y-%m-%d")
oneyearago = date.today() - relativedelta(years=1)
start_date = oneyearago.strftime("%Y-%m-%d")
df_discussions = df_discussions.loc[df_discussions['created'].between(start_date, end_date)]
nb_discussions = 0
nb_discussions_with_legit_answer = 0
time_to_answer = []
k = 0
for discussion_id in df_discussions['id']:
    discussion = requests.get(
        f"https://www.data.gouv.fr/api/1/discussions/{discussion_id}/"
    ).json()
```
Thoughts?

dgv/impact/DAG.py
```diff
@@ -0,0 +1,83 @@
...
+
+    download_history = BashOperator(
+        task_id='download_history',
+        bash_command=(
+            f"bash {AIRFLOW_DAG_HOME}{DAG_FOLDER}"
+            f"scripts/script_dl_history.sh {DATADIR} "
+        )
+    )
+
+    calculate_metrics = PythonOperator(
```
geoffreyaldebert, line 53: for clearer code, I think we can split this task into 4 tasks (actually only 2) for every kpis. Do you agree ?

dgv/impact/task_functions.py
```diff
@@ -0,0 +1,188 @@
...
+            'commentaires': 'les délais sont écrétés à 30 jours'
+        },
+    ]
+    df = pd.DataFrame(data)
+    history = pd.read_csv(f'{DATADIR}/history.csv')
+    final = pd.concat([df, history])
+    df.to_csv(os.path.join(DATADIR, "statistiques_impact_datagouvfr.csv"), index=False, encoding="utf8")
+
+
+def send_stats_to_minio():
```
geoffreyaldebert, line 146: can we duplicate saving file, one with this name, one with this name and the actual date. To keep an history in case there is some problem ?

geoffreyaldebert (changes requested):

dgv/impact/DAG.py
Pierlou: Done with the last commit ✅

dgv/impact/task_functions.py
Pierlou: Done with last commit ✅

PR #83, opened 2023-10-09
[DAG metrics] use headers to optimize processes

Pierlou:


dgv/monitoring/DAG-bizdev.py
```diff
@@ -58,9 +58,15 @@ async def crawl_reuses(reuses):
 
 
 def get_unavailable_reuses():
+    headers = {
+        "X-Fields": "data{id,url},next_page"
```
maudetes, line 62: `next_page` could be easily forgotten, since not used by the caller of èget_all_from_api_query`, I wonder if we should add it automatically in the function?

maudetes (approved):
Oh nice! :rocket:

I haven't checked all masks, I trust you on that :)

dgv/monitoring/DAG-bizdev.py
Pierlou: Could be! I had forgotten it in the first place but I soon realized because it throws an error, so maybe it's OK to leave it like that too, also because I'm not sure how to handle calls to others APIs (e.g metrics)

Pierlou:
```
if headers and not next_page in headers:
    headers = headers.replace('}', ',next_page}')
```
Or something like this? Is it worth it?

maudetes: I think it a bit specific if we're dealing with generic headers, but could work with an explicit mask param?

```
def get_all_from_api_query(..., mask=None):
    headers = {'X-fields': mask + f',{next_page}'} if mask else None
```

Pierlou: Done in the following commits ✅

geoffreyaldebert (approved):
nice

PR #91, opened 2023-10-18
[DAG DVF] refactor: truncate tables instead of dropping

Pierlou:


geoffreyaldebert (commented):
can you do a truncate if table exists and if not let creation of all properties ?

geoffreyaldebert (approved):

PR #95, opened 2023-10-23
[data_processing][rne]Download RNE stock files using ftp instead of https

HAEKADI:
INPI changed where RNE stock files are stored. They are no longer accessible from a web page but can only be downloaded via ftp.

### DIFF
--- a/config.py
+++ b/config.py
@@ -50,8 +50,10 @@ SECRET_INSEE_BEARER = Variable.get("SECRET_INSEE_BEARER", "")
 SECRET_INPI_USER = Variable.get("SECRET_INPI_USER", "")
 SECRET_INPI_PASSWORD = Variable.get("SECRET_INPI_PASSWORD", "")
 
+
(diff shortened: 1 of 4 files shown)


Pierlou (approved):
Looks good to me, but I'm curious: why did you switch to bash instead of python for this?

XavierJp (approved):
Nice. Is lftp always installed on debian distros ?

HAEKADI:
@Pierlou it is easier in my opinion and offers a more direct access since `lftp` is already installed with with the Docker file. Doable with `ftplib` on Python as well. Just a personal choice to mix it up ;)

HAEKADI:
@XavierJp No it's not, but it is installed in with docker in our data engineering stack for potential use https://github.com/etalab/data-engineering-stack/blob/d685a149fcae4261ddf659c3df3a80daa5b97111/Dockerfile#L21.

PR #97, opened 2023-10-24
[data-processing][rne] Unzip one stock rne file at a time

HAEKADI:
Unzipped files are about 100G, and the prod server isn't big enough. Instead of extracting all files at once, only one file is extracted at a time.


geoffreyaldebert (approved):

HAEKADI:
RNE data will be processed with annuaire workflows.

PR #108, opened 2023-11-10
Metrics clean matviews

geoffreyaldebert:
1. Change semantic for materialized views (exposed in api) from `visit` to `download` for resources
2. Remove `outlink` metrics (from matomo) for materialized views used in udata (because actually they are empty)


maudetes (approved):
:rocket:

PR #112, opened 2023-11-20
[DAG recommendations] use mask to get datasets ids

Pierlou:


maudetes (approved):
I haven't tested, but this seems fair

PR #130, opened 2023-12-08
[DAG météo] refactor avec balises

Pierlou:


data_processing/meteo/task_functions.py
```diff
@@ -29,44 +30,80 @@
...
+        _ = _.replace(f"{h}-", "")
+    return _
+
+
+def build_file_id(file, path):
+    # we are creating ids for files based on their name so that we can
+    # match files that have been updated with a name change
+    # for instance:
+    # QUOT_SIM2_latest-2020-202310.csv.gz becomes QUOT_SIM2_latest-2020-202311.csv.gz
+    # when Nov2023 is added, but it actually *is* just a content update of the same file
```
maudetes, line 48: Could we have an example of id for this file?

data_processing/meteo/task_functions.py
```diff
@@ -29,44 +30,80 @@
...
+    file_id = file
+    if config.get(path, {}).get("source_pattern"):
+        params = re.match(config[path]["source_pattern"], file)
+        params = params.groupdict() if params else {}
+        if "PERIOD" in params and any([
+            h in params["PERIOD"] for h in hooks
+        ]):
+            # this will have to change if more hooks are added
+            params["PERIOD"] = "latest" if "latest" in params["PERIOD"] else "previous"
```
maudetes, line 57: If we're mostly replacing `PERIOD`, we could use a regex to replace the period matched?

data_processing/meteo/task_functions.py
```diff
@@ -186,60 +264,113 @@ def upload_files_datagouv(ti, minio_folder):
...
+                        type="main" if not is_doc else "documentation",
+                        format=get_file_extention(file_with_ext),
+                        description=description,
+                    )
+                    new_files_datasets.add(path)
+                    updated_datasets.add(path)
+                # resource is updated with a new name, we redirect the existing resource, rename it
+                # and update size and description
+                # values because we are dealing with the resource's new name
+                elif clean_file_path in files_to_update_new_name.values():
```
maudetes, line 348: We should look in files_to_update_new_name keys, and not values, since keys are new names.

```suggestion
                elif clean_file_path in files_to_update_new_name:
```

maudetes (approved):
The logic seems good, well done for this tedious job! :clap: 
Having to deal with edge cases make the code quite complicated, but we can improve on it later on I think.

PR #142, opened 2023-12-19
[DAG bizdev] add user agents when crawling reuses

Pierlou:
and lint code

### DIFF
--- a/dgv/monitoring/DAG-bizdev.py
+++ b/dgv/monitoring/DAG-bizdev.py
@@ -228,7 +228,10 @@ def create_all_tables():
             datasets_visited[k].update({
                 'title': r.get('title', None),
                 'url': r.get('page', None),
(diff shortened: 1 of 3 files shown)


dgv/monitoring/DAG-bizdev.py
```diff
@@ -29,13 +29,38 @@
...
+        async with _method(
+            url,
+            timeout=15,
+            allow_redirects=True,
+            headers={"User-Agent": agents[idx]}
+        ) as r:
+            if r.status in [403, 405, 500]:
+                if idx < len(agents) - 1:
+                    # trying different user agents
+                    return await url_error(url, session, method=method, idx=idx + 1)
```
maudetes, line 60: I don't think we should spam reuses for each user agent if getting an error. Maybe we should just try with a random user agent from the list once?

maudetes (approved):
(not sure about digest notebooks that shows as `Empty file` in Files Changed)

PR #151, opened 2024-01-29
[DAG notification] add language to spam detection in moderation

Pierlou:


maudetes (commented):
Thank you! I think we should probably run this notebook more than once an hour.

PR #199, opened 2024-04-15
Show user website in digest

maudetes:
It's a good hint on spam accounts. Most legit users don't set a website and looking at the website gives a lot of info


Pierlou (approved):
Sounds like we're gonna have some crunchy stuff 👀

maudetes:
> Sounds like we're gonna have some crunchy stuff 👀

Yes, sadly :sweat_smile:

PR #230, opened 2024-05-29
Dag HVD

Pierlou:


dgv/monitoring/DAG-hvd.py
```diff
@@ -0,0 +1,185 @@
...
+
+def slugify(s):
+    return unidecode(s.lower().replace(" ", "-").replace("'", "-"))
+
+
+def get_hvd(ti):
+    print("Getting suivi ouverture")
+    ouverture_hvd_api = 'https://ouverture.data.gouv.fr/api/high_value_datasets'
+    df_ouverture = pd.DataFrame(requests.get(ouverture_hvd_api).json())
+    goal = len(df_ouverture)
```
maudetes, line 32: I think it's the number of *unique* hvd name? I mean we could compare the number of HVD "ensemble de données" for which we have at least one existing dataset.

maudetes (approved):
Thank you for this PR! :clap

I haven't checked all thoroughly but we can iterate on this later!

PR #286, opened 2024-07-30
[DAG notification] ping for unseen reports

Pierlou:


dgv/monitoring/DAG-notification.py
```diff
@@ -511,6 +548,11 @@ def publish_mattermost(ti):
         python_callable=alert_if_awaiting_spam_comments,
     )
 
+    alert_if_awaiting_spam_comments = PythonOperator(
```
maudetes, line 551: What was that not here?

dgv/monitoring/DAG-notification.py
```diff
@@ -183,6 +183,43 @@ def alert_if_awaiting_spam_comments():
         send_message(message, MATTERMOST_MODERATION_NOUVEAUTES)
 
 
+def alert_if_new_reports():
+    # DAG runs every 5min but if it fails we catch up with this variable
+    previous_report_check = Variable.get(
```
maudetes, line 188: Can't we fetch latest run instead of using variables?

maudetes (approved):
Looks good to me!

dgv/monitoring/DAG-notification.py
Pierlou: Forgot to modify after copy/paste 🙃 thanks for the alert

dgv/monitoring/DAG-notification.py
Pierlou: That's more complex, getting the latest run in which this specific task didn't fail. I can try if needed, but this should also work

PR #361, opened 2024-11-25
Extract site from user about markdown for digest

maudetes:
It is useful to monitor spam user to check the markdown `about` field on top of the `website` dedicated field.

Only deal with inline markdown (see [stackoverflow regex reference](https://stackoverflow.com/questions/30734682/extracting-url-and-anchor-text-from-markdown-using-python/30738268#30738268) used).

### DIFF
--- a/dgv/monitoring/utils.py
+++ b/dgv/monitoring/utils.py
@@ -1,3 +1,5 @@
+import re
+
 import requests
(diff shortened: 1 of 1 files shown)


Pierlou (approved):
🕵️

PR #375, opened 2024-12-23
[DAG metrics] Fix matomo catalog process

hacherix:
Make the column selection clear when dumping the dataframe in organization-outlinks.csv. The dataframe had two "id" columns that were conflicting. The bug arised following a dependency upgrade.

Also add retries when calling the Matomo API when it returns a 502. Currently 5 retries before raising an error.

Proof of fix:
image


Pierlou (approved):
Thanks for the quick fix!

PR #379, opened 2025-01-06
[DAG stats] create dag for datagouv visits

Pierlou:
To update https://www.data.gouv.fr/fr/datasets/statistiques-de-consultation-de-data-gouv-fr/

### DIFF
--- a/dgv/stats/DAG.py
+++ b/dgv/stats/DAG.py
@@ -1,0 +1,57 @@
+from datetime import timedelta, datetime
+from airflow.models import DAG
+from airflow.operators.bash import BashOperator
+from airflow.operators.python import PythonOperator
+
+from datagouvfr_data_pipelines.dgv.stats.task_functions import (
+    TMP_FOLDER,
+    DATADIR,
+    create_current_year_if_missing,
+    update_current_year,
+)
+
+DAG_NAME = 'dgv_stats'
+
+default_args = {
+    'retries': 5,
+    'retry_delay': timedelta(minutes=5),
+}
+
+with DAG(
+    dag_id=DAG_NAME,
+    schedule_interval='55 23 * * *',
+    start_date=datetime(2023, 10, 15),
+    catchup=False,
+    dagrun_timeout=timedelta(minutes=20),
+    tags=["datagouv", "stats", "metrics"],
+    default_args=default_args,
+) as dag:
+
+    clean_previous_outputs = BashOperator(
+        task_id="clean_previous_outputs",
+        bash_command=(
+            f"rm -rf {TMP_FOLDER} && "
+            f"mkdir -p {TMP_FOLDER} && "
+            f"mkdir -p {DATADIR}"
+        ),
+    )
+
+    create_current_year_if_missing = PythonOperator(
+        task_id='create_current_year_if_missing',
+        python_callable=create_current_year_if_missing,
+    )
+
+    update_current_year = PythonOperator(
+        task_id='update_current_year',
+        python_callable=update_current_year,
+    )
+
+    clean_up = BashOperator(
+        task_id="clean_up",
+        bash_command=f"rm -rf {TMP_FOLDER}",
+    )
+
+    create_current_year_if_missing.set_upstream(clean_previous_outputs)
+    update_current_year.set_upstream(clean_previous_outputs)
+    update_current_year.set_upstream(create_current_year_if_missing)
+    clean_up.set_upstream(update_current_year)
--- a/dgv/stats/DAG.py
+++ b/dgv/stats/DAG.py
@@ -31,7 +31,6 @@ with DAG(
         task_id="clean_previous_outputs",
         bash_command=(
             f"rm -rf {TMP_FOLDER} && "
-            f"mkdir -p {TMP_FOLDER} && "
             f"mkdir -p {DATADIR}"
         ),
     )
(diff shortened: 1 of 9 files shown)


dgv/stats/DAG.py
```diff
@@ -0,0 +1,57 @@
...
+    dagrun_timeout=timedelta(minutes=20),
+    tags=["datagouv", "stats", "metrics"],
+    default_args=default_args,
+) as dag:
+
+    clean_previous_outputs = BashOperator(
+        task_id="clean_previous_outputs",
+        bash_command=(
+            f"rm -rf {TMP_FOLDER} && "
+            f"mkdir -p {TMP_FOLDER} && "
```
hacherix, line 34: `f"mkdir -p {TMP_FOLDER} && "` can be removed since it is included in `f"mkdir -p {DATADIR}"`

dgv/stats/task_functions.py
```diff
@@ -0,0 +1,83 @@
...
+    resources = requests.get(
+        f"{DATAGOUV_URL}/api/1/datasets/{config[AIRFLOW_ENV]['dataset_id']}/",
+        headers={"X-fields": "resources{id,title}"},
+    ).json()["resources"]
+    return resources
+
+
+def create_current_year_if_missing():
+    resources = get_current_resources()
+    current_year = datetime.today().year
```
hacherix, line 33: I think we should process yesterday's year. On the 1st of January we want to process the whole previous year no?

dgv/stats/task_functions.py
```diff
@@ -0,0 +1,83 @@
...
+
+def get_months(site_id, year):
+    today = datetime.today().strftime("%Y-%m-%d")
+    params = {
+        "idSite": site_id,
+        "module": "API",
+        "method": "API.get",
+        "format": "json",
+        "period": "day",
+        "date": f"{year}-01-01,{today}",
```
hacherix, line 56: nit: The Matomo API accepts `today` as a parameters so we don't have to calculate it.
`"date": f"{year}-01-01,today"`

dgv/stats/DAG.py
```diff
@@ -0,0 +1,57 @@
...
+DAG_NAME = 'dgv_stats'
+
+default_args = {
+    'retries': 5,
+    'retry_delay': timedelta(minutes=5),
+}
+
+with DAG(
+    dag_id=DAG_NAME,
+    schedule_interval='55 23 * * *',
```
hacherix, line 22: It looks like there is a delay for daily reports: https://fr.matomo.org/faq/general/faq_41/
Would you it be ok to move the schedule to 03:00UTC?

dgv/stats/task_functions.py
```diff
@@ -0,0 +1,83 @@
...
+    current_year = datetime.today().year
+    current_year_resource_id = None
+    for resource in resources:
+        if str(current_year) in resource["title"]:
+            current_year_resource_id = resource["id"]
+            break
+    if not current_year_resource_id:
+        # this should not happen by construction but we never know
+        raise ValueError("Missing current year resource")
+    df = get_months(109, current_year)
```
hacherix, line 75: I see that 109 is used in other DAGs as well, what do you think about making it a global variable?

dgv/stats/task_functions.py
```diff
@@ -0,0 +1,83 @@
...
+    ).json()["resources"]
+    return resources
+
+
+def create_current_year_if_missing():
+    resources = get_current_resources()
+    current_year = datetime.today().year
+    for resource in resources:
+        if str(current_year) in resource["title"]:
+            return
```
hacherix, line 36: Won't we end up with a resoure including only the 1st of January data without updating it anymore for the rest of the year?

dgv/stats/DAG.py
Pierlou: True!

dgv/stats/task_functions.py
Pierlou: I changed the behaviour so that we update the dataset everyday, by adding the previous day, therefore running at the end of everyday (23:55) to add the data about the current day. (NB: the data for 23:55 to midnight will be added the next day anyway)

dgv/stats/task_functions.py
Pierlou: That's convenient, I'll use it :)

dgv/stats/DAG.py
Pierlou: Well spotted! I am wondering what we prefer between:
- keeping it like this for now, knowing that the more accurate data will be added on the next day
- moving the DAG to a few hours later and move everything to "the day before"

dgv/stats/task_functions.py
Pierlou: Could be indeed!

dgv/stats/task_functions.py
hacherix: Those 5 minutes won't be backfill for the 31th of December no? But maybe we are good with it 😄

If so, can we use `datetime.utcnow().year` instead? I guess the Airflow server is UTC but at least we are sure it will be compatible with whatever local timezone a computer or server use.

dgv/stats/DAG.py
hacherix: I would vote for the second option so we directly send complete data no? And it would also solve the issue we are speaking in the above thread. But it is up to you, I will follow your lead!

Pierlou: You're right, this and the fact that we'd be missing the last 5min of the year are two good arguments, I'll change the behaviour :)

dgv/stats/task_functions.py
hacherix: True! 💪

hacherix (approved):
Looks great 🚀

PR #380, opened 2025-01-07
DAG contrôle sanitaire de l'eau

Pierlou:
Also created an util function to check if a dataset has been updated more recently than a specific resource (a syntax used in this DAG and two others) and refactored accordingly

### DIFF
--- a/data_processing/sante/controle_sanitaire_eau/config/dgv.json
+++ b/data_processing/sante/controle_sanitaire_eau/config/dgv.json
@@ -1,0 +1,68 @@
+{
+    "RESULT": {
+        "csv" : {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "b1b82209-93c4-4e51-af26-a9a404ba1fe2"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        },
+        "parquet" : {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "97cf8671-ec7f-4e33-a108-d22c33107970"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        }
+    },
+    "COM_UDI": {
+        "csv" : {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "97c7e4ee-d5be-4709-b255-c4506aa42bff"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        },
+        "parquet": {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "fffd9023-107c-4ed3-a187-e2038a9c35ff"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        }
+    },
+    "PLV": {
+        "csv" : {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "2a0605b7-72dd-4529-a0d4-73a471a5cc87"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        },
+        "parquet": {
+            "dev": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": "cacb4bf4-7bff-4c57-8c62-850bdc37be83"
+            },
+            "prod": {
+                "dataset_id": "5cf8d9ed8b4c4110294c841d",
+                "resource_id": ""
+            }
+        }
+    }
+}
--- a/data_processing/sante/controle_sanitaire_eau/config/dgv.json
+++ b/data_processing/sante/controle_sanitaire_eau/config/dgv.json
@@ -1,6 +1,6 @@
 {
     "RESULT": {
-        "csv" : {
+        "csv.gz" : {
             "dev": {
                 "dataset_id": "5cf8d9ed8b4c4110294c841d",
                 "resource_id": "b1b82209-93c4-4e51-af26-a9a404ba1fe2"
@@ -19,6 +19,9 @@
                 "dataset_id": "5cf8d9ed8b4c4110294c841d",
                 "resource_id": ""
             }
+        },
+        "dtype": {
+            "cdparametre": "INT32"
         }
     },
     "COM_UDI": {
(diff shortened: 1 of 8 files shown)


data_processing/sante/controle_sanitaire_eau/task_functions.py
```diff
@@ -0,0 +1,147 @@
...
+def send_to_minio(file_type):
+    minio_open.send_files(
+        list_files=[
+            {
+                "source_path": f"{DATADIR}/",
+                "source_name": f"{file_type}.{ext}",
+                "dest_path": "controle_sanitaire_eau/",
+                "dest_name": f"{file_type}.{ext}",
+            }
+            for ext in ["csv" if file_type != "RESULT" else "csv.gz", "parquet"]
```
hacherix, line 107: nit (so feel free to ignore 😄): To not hardcode `RESULT` use case and the `ext` list, I think the `ext` variable could be provided as kwargs like the `file_type` variable.
Same in `publish_on_datagouv()`. We would have to exclude `dtype` instead.

data_processing/sante/controle_sanitaire_eau/task_functions.py
```diff
@@ -0,0 +1,147 @@
...
+        csv_to_parquet(
+            f"{DATADIR}/{file_type}.csv",
+            sep=',',
+            dtype={
+                # specific dtypes are listed in the config, default to str
+                c: config[file_type]["dtype"].get(c, "VARCHAR")
+                for c in columns[file_type]
+            },
+        )
+        if file_type == "RESULT":
```
hacherix, line 93: To make it more modular, we could replace it by:
```
if config[file_type].get("csv.gz", ""): 
```

data_processing/sante/controle_sanitaire_eau/task_functions.py
```diff
@@ -0,0 +1,147 @@
...
+    for idx, resource in enumerate(resources):
+        print(resource["title"])
+        year = int(resource["title"].split(".")[0].split("-")[1])
+        r = requests.get(resource["url"])
+        r.raise_for_status()
+        with ZipFile(BytesIO(r.content)) as zip_ref:
+            for _, file in enumerate(zip_ref.namelist()):
+                print(">", file)
+                file_type = "_".join(file.split("_")[1:-1])
+                assert file_type in config.keys()
```
hacherix, line 61: Love your CSV conformity checks! 🚀

data_processing/sante/controle_sanitaire_eau/task_functions.py
```diff
@@ -0,0 +1,147 @@
...
+                    print(columns[file_type])
+                    print(list(df.columns))
+                    raise ValueError('Columns differ between files')
+                df["annee"] = year
+                df.to_csv(
+                    f"{DATADIR}/{file_type}.csv",
+                    index=False,
+                    encoding="utf8",
+                    mode="w" if idx == 0 else "a",
+                    header=idx == 0,
```
hacherix, line 80: Shouldn't it be the index of `enumerate(zip_ref.namelist())` instead of `enumerate(resources)`? 🤔

data_processing/sante/controle_sanitaire_eau/config/dgv.json
```diff
@@ -7,7 +7,7 @@
             },
             "prod": {
                 "dataset_id": "5cf8d9ed8b4c4110294c841d",
-                "resource_id": ""
+                "resource_id": "fee50ac3-df62-4b38-aaf5-4831c6ba65c7"
```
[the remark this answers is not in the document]
Pierlou: It's very not intuitive, it's because it's a community resource. This could change, we'll open a ticket!

utils/datagouv.py
```diff
@@ -566,3 +566,26 @@ def check_duplicated_orga(slug: str) -> tuple[bool, Optional[str]]:
...
+    resources = datagouv_session.get(
+        f"https://{prefix}.data.gouv.fr/api/1/datasets/{dataset_id}/",
+        headers={"X-fields": "resources{internal{last_modified_internal}}"}
+    ).json()['resources']
+    lastest_update = datagouv_session.get(
+        f"https://{prefix}.data.gouv.fr/api/2/datasets/resources/{reference_resource_id}/",
+        headers={"X-fields": "internal{last_modified_internal}"}
+    ).json()["internal"]["last_modified_internal"]
+    return any(
+        r["internal"]["last_modified_internal"] > lastest_update for r in resources
```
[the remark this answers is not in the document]
Pierlou: It's the most accurate date metadata we have in this case (dates in udata are a **big** thing 👀)

utils/datagouv.py
```diff
@@ -566,3 +566,26 @@ def check_duplicated_orga(slug: str) -> tuple[bool, Optional[str]]:
...
+def check_if_recent_update(
+    reference_resource_id: str,
+    dataset_id: str,
+    on_demo: bool = False,
+) -> bool:
+    """
+    Checks whether any resource of the specified dataset has been update more recently
+    than the specified resource
+    """
+    prefix = "demo" if on_demo else "www"
```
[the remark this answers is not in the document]
Pierlou: It's true that we never (so far) want to target demo in prod, but we often want to target prod from DAGs even in the dev process (to get fresh data). This syntax (not the cleanest I admit) allows to choose which data.gouv env you want to target, no matter which Airflow env you come from.

data_processing/sante/controle_sanitaire_eau/task_functions.py
Pierlou: I don't mind having things hardcoded here as the structure most likely won't change, it's really only for this file out of the three

data_processing/sante/controle_sanitaire_eau/task_functions.py
Pierlou: Same here, this exception is not likely to change, so hardcoding it feels fine I'd say

data_processing/sante/controle_sanitaire_eau/task_functions.py
Pierlou: I don't think so, `enumerate(resources)` gives the index for which group of files we are processing, and whether or not to write the header/append instead of erase depends on it, whereas `enumerate(zip_ref.namelist())` gives the index of the file within its group (and I don't know why there's an enumerate here as we don't use the index btw 🤔)

utils/datagouv.py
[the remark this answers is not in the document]
hacherix: Got it! By default we target `prod` here. Could we do it the other way around to be safe?
Like by default we use `AIRFLOW_ENV` to check if we target demo or not. But with an option to force the use of prod even in dev?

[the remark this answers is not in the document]
Pierlou: I'd say it's fair to target prod by default here, because we usually want to refresh data depending on the freshest platform (demo is far behind) (and tbh in dev mode I usually just add a `return True` in this task to trigger the following ones, where the processing really happens)

data_processing/sante/controle_sanitaire_eau/task_functions.py
hacherix: True! The resources are split per year and not per type my bad! Nice one!

hacherix (approved):

PR #381, opened 2025-01-08
[DAG météo FTP] sort resources if new ones have been added

Pierlou:
:warning: Do not merge before we're sure that sorting resources is not the cause of resource duplication


data_processing/meteo/ftp_processing/task_functions.py
```diff
@@ -437,7 +438,24 @@ def handle_updated_files_new_name(ti, minio_folder):
     ti.xcom_push(key="new_files", value=new_files)
 
 
-def update_temporal_coverages(ti):
+def meteo_sort(resources, path):
+    keys = [
+        {
+            'keys': (
+                r['title'].split('_')[2],
+                r['title'].split('_')[4],
```
maudetes, line 446: I think it does not work for a documentation resource like https://www.data.gouv.fr/fr/datasets/donnees-climatologiques-de-base-decadaires/#/resources/20a94897-a3c2-4ac4-ba70-b13e2607f7a9

utils/datagouv.py
```diff
@@ -548,6 +548,32 @@ def post_comment_on_dataset(dataset_id: str, title: str, comment: str) -> reques
...
+    # sort_func is required to take a list of resources and
+    # return a list of dicts [{'id': resource_id}, ...] in the desired order
+    resources = datagouv_session.get(
+        f"{DATAGOUV_URL}/api/1/datasets/{dataset_id}/",
+    ).json()['resources']
+    sorted_ids = sort_func(resources)
+    if len(sorted_ids) != len(resources):
+        raise ValueError("The sorted list has a different number of elements, aborting")
+    if len(sorted_ids) != len(set(r["id"] for r in sorted_ids)):
+        raise ValueError("An id has been duplicated in the sort process", resources, "vs", sorted_ids)
```
maudetes, line 568: :pray:

maudetes (commented):
I don't see what would go wrong, it does seem to makes sense

data_processing/meteo/ftp_processing/task_functions.py
Pierlou: Thanks for the catch!

PR #395, opened 2025-01-14
[DAG recommendations] refactor

Pierlou:
A few questions because I am not familiar with the big picture:
- how is [the recommendations file from schema.data.gouv](https://github.com/datagouv/schema.data.gouv.fr/blob/main/site/.vuepress/public/api/recommendations.json) used by `udata`?
- couldn't we delete the folder with saved notebooks on Minio (at `.../data-pipeline-open/schema%2Fschemas_recommendations%2F`)? There are no displays in the notebook, it's litterally just the code


nicolaskempf57:
I don't have access to the deployment pipeline but the file should be added as recommendation sources. You can try with a local udata by changing your udata.cfg, for example mine is :

```
RECOMMENDATIONS_SOURCES = {
    'edito': 'https://etalab.github.io/datagouvfr-recommendations-edito/recommendations.json',
    'matomo': 'https://etalab.github.io/piwik-covisits/recommendations.json'
}
```

schema/recommendations/task_functions.py
```diff
@@ -0,0 +1,77 @@
...
+        ).json()["slug"]
+        print(
+            f"Working on schema {schema_id}, consolidated at {consolidated_dataset_id}"
+        )
+        dataset_ids = datasets_for_schema(schema_id)
+        print(f"Found {len(dataset_ids)} associated with schema {schema_id}")
+        recommendations.extend([
+            build_recommendation(consolidated_slug, d) for d in dataset_ids
+        ])
+    # to make sure that there is not more than one recommendation per dataset (do we want this?)
```
maudetes, line 68: I don't think it matters too much if there are two recommendations? If only one should be displayed?

maudetes (approved):
Thank you! Easy to follow, well done :clap:

Do you want to run it and make a diff with the previous and see if some things differ but shouldn't?

schema/recommendations/task_functions.py
Pierlou: Same, I was wondering if we wanted to keep more than one, and if so what we should display, because in a few weeks we may have multiple schemas referenced in single datasets (due to https://schema.data.gouv.fr/etalab/schema-indice-durabilite/)

Pierlou:
[Here](https://github.com/datagouv/schema.data.gouv.fr/pull/329) is an overview of the changes that would be pushed (quite a bit because the DAG wasn't doing its job for a while). One main difference: we now use the datasets' slugs instead of ids (because slugs are used [here](https://etalab.github.io/datagouvfr-recommendations-edito/recommendations.json), not sure it matters but for consistency)

hacherix (approved):

schema/recommendations/task_functions.py
Pierlou: I removed the part that removed multiple recommendations for single datasets, we'll see how we want to handle it in the front 🚀

PR #397, opened 2025-01-21
[DAG HVD] update grist table

Pierlou:


dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+            or row["format"] in ["ogc:wms", "ogc:wfs", "wms", "wfs"]
+        ):
+            # "fake" resources that are actually dataservices
+            contact_point = requests.get(
+                f"https://www.data.gouv.fr/api/1/datasets/{dataset_id}/"
+            ).json()["contact_point"] or {}
+            return (
+                row["title"],
+                url,
+                url,
```
maudetes, line 263: I think that only URL with `request=getcapabilities` would be considered ok for documentation (or endpoint description)

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+def get_hvd_category_from_tags(tags):
+    tags = tags.split(",")
+    for tag in tags:
+        if tag in HVD_CATEGORIES:
+            return tag
+
+
+def dataservice_information(dataset_id, df_dataservices, df_resources):
```
maudetes, line 232: Could you maybe add a docstring to explicit the 5 different return values?

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    "observation-de-la-terre-et-environnement",
+    "statistiques",
+]
+API_TABULAIRE_ID = '673b0e6774a23d9eac2af8ce'
+
+
+def get_hvd_ouverture(df_ouverture, url):
+    df_res = df_ouverture.loc[df_ouverture["URL_Telechargement"] == url]
+    if len(df_res):
+        return df_res.iloc[0]["Ensemble_de_donnees"]
```
maudetes, line 222: I think, info will disappear since this is replacing the existing ouverture table.
Let's keep it for now :)

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    df_datasets['hvd_category_datagouv'] = df_datasets["tags"].apply(get_hvd_category_from_tags)
+    df_datasets.rename({"license": "license_datagouv"}, axis=1, inplace=True)
+    print("Processing...")
+    (
+        df_datasets['api_title_datagouv'],
+        df_datasets['endpoint_url_datagouv'],
+        df_datasets['endpoint_description_datagouv'],
+        df_datasets['api_web_datagouv'],
+        df_datasets['contact_point_datagouv']
+    ) = zip(*df_datasets["id"].apply(
```
maudetes, line 320: I think the zip approach comes from my notebook. Out of curiosity, is this how you would do as well to assign multiple columns?

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    df_datasets.to_csv(DATADIR + "fresh_hvd_metadata.csv", index=False)
+
+
+def update_grist():
+    old_hvd_metadata = get_table_as_df(
+        doc_id=DOC_ID,
+        table_id="Hvd_metadata_res",
+        columns_labels=False,
+    )
+    if old_hvd_metadata["id2"].nunique() != len(old_hvd_metadata):
```
maudetes, line 332: ah should we rename the column `id` in grist? Or is it protected and can't be used as a column technical name?

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    fresh_hvd_metadata = pd.read_csv(DATADIR + "fresh_hvd_metadata.csv").rename(
+        # because the "id" column in grist has the identifier "id2"
+        {"id": "id2"},
+        axis=1
+    )
+    if fresh_hvd_metadata["id2"].nunique() != len(fresh_hvd_metadata):
+        raise ValueError("New table has duplicated dataset ids")
+    removed_hvd = (set(old_hvd_metadata["id2"]) - set(fresh_hvd_metadata["id2"]))
+    for hvd_id in removed_hvd:
+        send_message(
```
maudetes, line 343: :ok_hand: thank you!

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    updates = 0
+    for dataset_id in old_hvd_metadata["id2"]:
+        row_old = old_hvd_metadata.loc[old_hvd_metadata["id2"] == dataset_id].iloc[0]
+        row_new = fresh_hvd_metadata.loc[fresh_hvd_metadata["id2"] == dataset_id].iloc[0]
+        new_values = {}
+        for col in columns_to_update:
+            if (
+                (isinstance(row_new[col], str) and row_new[col])
+                or (isinstance(row_new[col], float) and not isnan(row_new[col]))
+            ) and row_old[col] != row_new[col]:
```
maudetes, line 368: I am unclear about all the conditions here

dgv/monitoring/hvd/task_functions.py
```diff
@@ -0,0 +1,395 @@
...
+    if fresh_hvd_metadata["id2"].nunique() != len(fresh_hvd_metadata):
+        raise ValueError("New table has duplicated dataset ids")
+    removed_hvd = (set(old_hvd_metadata["id2"]) - set(fresh_hvd_metadata["id2"]))
+    for hvd_id in removed_hvd:
+        send_message(
+            f":alert: [Ce jeu de données](https://www.data.gouv.fr/fr/datasets/{hvd_id}/)"
+            " a perdu son tag HVD @clarisse",
+            MATTERMOST_MODERATION_NOUVEAUTES
+        )
+    columns_to_update = [
```
maudetes, line 348: Shouldn't it be all columns except id?

maudetes (approved):
Thank you! The code looks nice :star_struck: 

I kept my review to the grist table only :)

dgv/monitoring/hvd/task_functions.py
Pierlou: Mmhh should we change the conditions then? Because for instance in [this dataset](https://www.data.gouv.fr/fr/datasets/masses-deau-cours-deau-guadeloupe-version-rapportage-2010/) we currently get the resource that doesn't have the whole link

dgv/monitoring/hvd/task_functions.py
Pierlou: Not sure I understand, should we change this?

dgv/monitoring/hvd/task_functions.py
Pierlou: There is [a different solution](https://stackoverflow.com/questions/23586510/return-multiple-columns-from-pandas-apply) (and I'm pretty sure there are many more), not sure what's best but for such a small dataframe it shouldn't be a problem I guess

dgv/monitoring/hvd/task_functions.py
Pierlou: I don't think so, our Grist utils should work fine with labels/ids, it's OK to keep it so

dgv/monitoring/hvd/task_functions.py
Pierlou: We want to replace the current value if
- the new cell content is a not-empty string OR the new cell is not NaN
- AND the new value is different from the current one

Is that fine?

dgv/monitoring/hvd/task_functions.py
Pierlou: I don't know, all the `dataeuropa` columns are not retrieved any more, I checked which columns we have here compared to the ones we have in the dataset, and added this as a safeguard to make sure we only erased known columns, but we can change this

dgv/monitoring/hvd/task_functions.py
maudetes: not yet :)

dgv/monitoring/hvd/task_functions.py
maudetes: Why bother about being empty or NaN? Is it because grist will convert it and we will see it differ when it's not actually the case?

Pierlou: To prevent updating with unwanted values (like replacing an empty string with NaN); we only update if the new content is clearly new info (and different from the current one)

PR #369, opened 2024-12-18
[DAG metrics] add dataservices metrics

hacherix:
Ajout des dataservices aux mesures calculées dans le DAG `metrics` à partir des logs HAProxy.
De nouvelles tables et vues sont disponibles dans la BDD Postgres pour accueillir ces données. Sur le même modèle que les anciennes.

De mesures avec une plus faible granularité sur l'usage des datasets/resources/etc. sont ajoutées.
* `nb_visit_api1`
* `nb_visit_api2`
* `nb_visit_apis` (somme de api1 et api2)
* `nb_visit_fr`
* `nb_visit_en`
* `nb_visit_es`
* `nb_visit_total`
* La mesure d'origine `nb_visit` qui ne concerne que le site reste identique.
`nb_visit` correspond donc à la somme de `nb_visit_fr`, `nb_visit_es` et `nb_visit_en`.

Le code a été réusiné pour le rendre plus modulaire et quelques tests ont été ajoutés sur des parties critiques du code.

Aucun outlinks calculé pour les dataservices car ce ne semble pas s'y prêter. Le swagger est affiché directement sur la page. Mais cela peut toujours être modifié au besoin.

### Impact

Changement sur les metrics sur le 22 janvier 2025 :

| Objet | Nb IDs (ancien) | Nb IDs (nouveau) | Nb visites (ancien) | Nb visites (nouveau) |
|---|---|---|---|---|
| datasets | 67 929 | 73 089 | 191 514 | 298 239 |
| organisations | 5 209 | 5 198 | 11 450 | 24 010 |
| resources | 44 943 | 62 824 | 7 869 775 | 7 919 893 |
| reuses | 4 206 | 4 305 | 10 160 | 39 479 |
| dataservices | - | - | 343 | 4 061 |

### Bug fix

- Les logs d'URL complètes n'étaient pas comptabilisés précédemment.

`https://www.data.gouv.fr/fr/datasets/obsolete-le-calendrier-scolaire-2019-2020-et-2020-2021-format-ical/` correspond maintenant à une visite pour le dataset correspondant. L'impact sur les chiffres peut être important. Par exemple, pour le dataset en exemple on passe de 8 visites à 284 sur la journée du 22 janvier.

- Empêche la duplication des metrics sur une journée dans le cas où un DAG est exécuté une seconde fois.

- Fix où les ressources téléchargées à partir de l'ID et non du lien static n'étaient pas comptabilisées lorsqu'un lien static existe.

### Breaking Changes

[...]

PR #419, opened 2025-02-07
[DGV.METRICS] Add matomo database duplication safety + bug fixes

hacherix:
* Add Matomo duplication safety in case we run the DAG twice the same day
* Revert logic for the static resources visits + add comments so we understand the reason behind it
  Related to the discussion that was going on here: https://github.com/datagouv/datagouvfr_data_pipelines/pull/369#discussion_r1910079365 
* Fix the `return_sql_results()` error when there the cursor has nothing to fetch
```bash
Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 192, in execute
    return_value = self.execute_callable()
  File "/home/airflow/.local/lib/python3.9/site-packages/airflow/operators/python.py", line 209, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
  File "/opt/airflow/dags/datagouvfr_data_pipelines/dgv/metrics/task.py", line 39, in create_metrics_tables
    pgtool.execute_sql_file(
  File "/opt/airflow/dags/datagouvfr_data_pipelines/utils/postgres.py", line 48, in execute_sql_file
    data = self._return_sql_results(cur)
  File "/opt/airflow/dags/datagouvfr_data_pipelines/utils/postgres.py", line 76, in _return_sql_results
    data = cur.fetchall()
psycopg2.ProgrammingError: no results to fetch
```
* Add more tests and explanations for the log info unit test


utils/postgres.py
```diff
@@ -35,21 +34,30 @@ def raise_if_not_file(self, file: File):
...
-            data = return_sql_results(cur)
+            data = self._return_sql_results(cur)
             self.conn.commit()
         return data
 
-    def execute_sql_file(self, file: File) -> list[dict]:
+    def execute_sql_file(
+        self,
+        file: File,
+        replacements: dict[str, str] = {},
```
Pierlou, line 44: Nice addition!

utils/postgres.py
```diff
@@ -59,26 +67,26 @@ def copy_file(self, file: File, table: str, has_header: bool) -> list[dict]:
...
-        data = cur.fetchall()
-    except psycopg2.ProgrammingError as e:
-        print(e)
-        data = None
-    if data:
-        columns = [desc[0] for desc in cur.description]
-        return [{k: v for k, v in zip(columns, d)} for d in data]
-    return []
+        Returns:
+                dict, None: result of sql query or None
```
Pierlou, line 82: Actually now only returns a (potentially empty) list
```suggestion
                list: result of sql query, or an empty list if no result
```

Pierlou (approved):
Thanks for the improvements 🚀

dgv/metrics/task.py
```diff
@@ -137,23 +138,38 @@ def aggregate_log(ti) -> None:
...
-                    )
+                catalog_dict: dict[str, str] = defaultdict()
+                static_uri = "https://static.data.gouv.fr/resources/"
+
+                # Resource catalog has no slug column but static
+                # URLs starting with https://static.data.gouv.fr/resources/$SLUG
+                # Using a resource ID will trigger a redirect to its static URL so we want:
+                # 1. All the slugs from the static URLs
+                df_slugs = (
+                    df_catalog.loc[lambda df: df["url"].str.contains(static_uri)]
```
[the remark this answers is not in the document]
hacherix: The use of the `lambda` is also very useful when method chaining. Since the underlying DataFrame change we can't directly use it as a reference.

utils/postgres.py
hacherix: Good catch! We both fixed the issue `psycopg2.ProgrammingError` but my rebase was incomplete sorry! I like your solution to return an empty list instead of None. Really clean!

dgv/metrics/task.py
```diff
@@ -137,23 +138,38 @@ def aggregate_log(ti) -> None:
...
-                # A few resource_id are common to multiple datasets
-                # Deduplicate them with priority to "dataset.archived" as False
-                # Otherwise keep the last one created
-                df_catalog = df_catalog.sort_values(
-                    by=["dataset.archived", "created_at"],
-                    ascending=[True, False]
+                catalog_dict.update(df_slugs.set_index("slug")["id"].to_dict())
+
+                # 2. All the IDs that don't have any static URL
+                #  Note: a few resource_id are common to multiple datasets
```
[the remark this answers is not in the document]
hacherix: I ran this query on the resources catalog:
```sql
select id
     , count(distinct "dataset.id") as n_datasets
     , count(distinct case when "dataset.archived" is not null then "dataset.id" end) as n_datasets_archived
from read_csv_auto('catalog_resources.csv')
group by 1
having n_datasets > 1
order by n_datasets desc;
```
Ce sont tous des datasets archivés.
[output.csv](https://github.com/user-attachments/files/18731584/output.csv)

PR #434, opened 2025-02-18
[Meta DAG] allow groups of DAGs in watchlist

Pierlou:


meta/task_functions.py
```diff
@@ -13,15 +14,30 @@
...
+    config = json.load(f)
+
+
+def get_ids(config: dict):
+    dags = http_hook.run('api/v1/dags').json()["dags"]
+    ids = []
+    for raw_id, included in config.items():
+        if not included:
+            continue
+        if raw_id.endswith("*"):
```
hacherix, line 29: Nice one! 🙌

PR #437, opened 2025-02-20
[IRVE] Investigate dag slowdown

hacherix:


data_processing/irve/geo_utils/geo.py
```diff
@@ -144,7 +148,7 @@ def enrich_row_address(row: pd.Series, session) -> pd.Series:
             # Lat lon do not match any commune
             enrich_row_address.no_match_coords += 1
 
-        if str(row[code_insee_col]) in row[address_col]:
+        if row[code_insee_col] and str(row[code_insee_col]) in row[address_col]:
```
Pierlou, line 151: I'm not sure what's supposed to be in the column at this point but if there can be `NaN` values this could be problematic, as `bool(NaN)` is True I think

data_processing/irve/geo_utils/geo.py
```diff
@@ -194,55 +198,63 @@ def enrich_row_address(row: pd.Series, session) -> pd.Series:
...
+    logging.info(
         "No indication of postcode/code INSEE in address or coordinates matching code INSEE field. "
         f"No enriching performed: {enrich_row_address.nothing_matches}/{total_rows}"
     )
     return df
 
 
-def improve_geo_data_quality(file_cols_mapping: Dict[str, Dict[str, str]]) -> None:
+def improve_geo_data_quality(
+        file_cols_mapping: dict[str, dict[str, str]]
```
Pierlou, line 238:
```suggestion
        file_cols_mapping: dict[str, dict[str, str]],
```

utils/schema.py
```diff
@@ -31,15 +33,14 @@
...
-# DEV : for local dev in order not to mess up with production
-# DATAGOUV_URL = 'https://www.data.gouv.fr'
+tqdm.pandas(desc="pandas progress bar", mininterval=30)
 
 VALIDATA_BASE_URL = (
     VALIDATA_BASE_URL + "/validate?schema={schema_url}&url={rurl}"
 )
 MINIMUM_VALID_RESOURCES_TO_CONSOLIDATE = 5
-api_url = f"{DATAGOUV_URL}/api/1/"
+# Target the prod website in any case since schemas are not available on the demo API
```
Pierlou, line 42: I'm not sure what you mean here?

Pierlou (commented):
Thanks for all the cleaning 🙏

data_processing/irve/geo_utils/geo.py
hacherix: Good question, I will check more in details. The idea here is to stop calling the geo api when `row[code_insee_col]` value is an empty string such as:
https://geo.api.gouv.fr/communes?codePostal=&fields=code,nom

utils/schema.py
hacherix: https://www.demo.data.gouv.fr/api/1/datasets/?schema=etalab/schema-irve-statique
Does not work.
So I just removed this comment :
```
# DEV : for local dev in order not to mess up with production
# DATAGOUV_URL = 'https://www.data.gouv.fr'
```
And just made the `api_url` use production.
My understanding was that to test in dev we need to uncomment `DATAGOUV_URL`.

It should be noted that the `upload_consolidated_irve` task is still upload to demo as far as I can see.

Did I get that wrong?

data_processing/irve/geo_utils/geo.py
hacherix: I went with something more explicit! Thanks!

utils/schema.py
Pierlou: It works but the URL has to be encoded: `https://demo.data.gouv.fr/api/1/datasets/?schema=etalab%2Fschema-irve-statique`

data_processing/irve/task_functions.py
```diff
@@ -377,53 +377,44 @@ def create_consolidation_reports_irve(
...
-    )
+    final_directory_clean_up(tmp_folder, output_data_folder)
 
 
 def upload_minio_irve(
-    TMP_FOLDER,
-    MINIO_BUCKET_DATA_PIPELINE_OPEN,
-    minio_output_filepath
-):
+    tmp_folder: Path,
```
Pierlou, line 397: You have to modify the `op_kwargs` of the task if you want to modify this don't you?

utils/schema.py
```diff
@@ -362,16 +331,12 @@ def is_validata_valid(rurl, schema_url, resource_api_url, validata_base_url=VALI
         report = make_validata_report(rurl, schema_url, resource_api_url, validata_base_url)
         try:
             res = report["report"]["valid"]
-        except:
-            print(
-                f"---- 🔴 No info in validata report for resource: {rurl}"
-            )
+        except KeyError:
+            logging.info(f"---- 🔴 No info in validata report for resource: {rurl}")
```
Pierlou, line 335: Maybe for these kinds of logs we could go for `warning` (but this can be for another PR)

hacherix: Good point! I am adding it to the ones I think make sense

data_processing/irve/task_functions.py
hacherix: True! Good catch!

utils/schema.py
hacherix: I just add a replace of `/` by `%2F` where this was causing issue. `%2F` works in production as well so now it should work for both. Thanks for the info!

Pierlou (approved):
Here we go I guess, thanks you for the quick work!

PR #438, opened 2025-02-24
[DAG IRVE] Get existing infos

Pierlou:


data_processing/irve/geo_utils/geo.py
```diff
@@ -110,7 +100,9 @@ def fix_code_insee(
     """
 
     @simple_connection_retry
-    def enrich_row_address(row: pd.Series, session) -> pd.Series:
+    def enrich_row_address(row: pd.Series, session: requests.Session) -> pd.Series:
+        if isinstance(row[code_insee_col], str) and row[code_insee_col]:
+            return row
```
hacherix, line 105: Should we `return row` directly only when `consolidated_is_code_insee_verified` so we still enrich new ones that may have a `code_insee_col` value?

Pierlou: True, thanks for the catch!

hacherix (approved):

hacherix (approved):
🚀

PR #442, opened 2025-03-03
[DAG bizdev] split into tasks and refactor

Pierlou:


hacherix (approved):
LGTM 🚀 
Could you give me access to the grist document where the data is dumped please? So I can look at it as well 🙏

PR #462, opened 2025-03-18
Fix digests and formation DAGs

Pierlou:
I am not sure about the `if x == x` 🤔


hacherix (approved):
Nice!!

The `val == val` is maybe for filtering out the np.nan? Not clear indeed. 🤔

Pierlou:
You're right, I'll fix this before merging

PR #473, opened 2025-04-11
Add DAGs check workflow

Pierlou:


.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+          name: Clone stack
+          command: git clone https://github.com/datagouv/data-engineering-stack.git
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
+  check-dags:
+    docker:
+      - image: apache/airflow:2.10.5-python3.12
+    steps:
```
hacherix, line 19: The `- checkout` step is missing I think.

PR #491, opened 2025-05-19
[DAGS siren publications] feat: restore checksums

Pierlou:


maudetes (approved):
Thank you for restoring the checksums

PR #514, opened 2025-07-02
generate tags for simplifions search

JeSuisUnCaillou:


dgv/simplifions/task_functions.py
```diff
@@ -59,6 +58,16 @@ def cleaned_row_with_subdata(row):
     }
     return formatted_row
 
+def generated_search_tags(topic):
+    tags = []
+    for attribute in ATTRIBUTES_FOR_TAGS:
+        for value in topic[attribute]:
+            tags.append(f'simplifions-{attribute}-{value}')
+    return tags
```
Pierlou, line 66: Nit
```suggestion
    return [
        f"simplifions-{attribute}-{value}"
        for attribute in ATTRIBUTES_FOR_TAGS
        for value in topic[attribute]
    ]
```

dgv/simplifions/task_functions.py
```diff
@@ -88,6 +98,7 @@ def update_topics(ti):
                 + "&".join([f"tag={tag}" for tag in simplifions_tags])
             ),
         )
+        if extras_nested_key in topic["extras"]
```
Pierlou, line 101: We expect to have external topics with this request on the tags?

dgv/simplifions/task_functions.py
JeSuisUnCaillou: Je trouve ça difficilement lisible perso cette sytaxe python avec les for après.
Jusque là j'y arrivais quand même, mais là avec 2 for imbriqués je suis un peu flabergasted : Le deuxième est exécuté avant le premier ? Ou est-ce que tu as inversé les deux ?

dgv/simplifions/task_functions.py
JeSuisUnCaillou: No, but when I changed the extra_nested_key, it raised an error because it couldn't find it in the extras of the topics fetched with `get_all_from_api_query` (because they had the former key)

Pierlou: Wasn't it the expected behaviour? If we're changing the nested key I'd say the good practice would be to delete the old topics and start fresh again, otherwise we'll keep the previous version of the topics (with the previous nested key), which will most likely come back to haunt us at some point

JeSuisUnCaillou:
> ah mais attend ça marche pas mon truc lol

Ah si. Fausse alerte.

PR #515, opened 2025-07-04
Simplifions DAG creates solutions topics

JeSuisUnCaillou:


dgv/simplifions/delete_demo_topics.py
Pierlou, line 1: Not sure why this is here, it a script you use locally right?

dgv/simplifions/task_functions.py
```diff
@@ -43,105 +55,175 @@ def clean_row(row):
...
 def generated_search_tags(topic):
-    return [
-        f"simplifions-{attribute}-{value}"
-        for attribute in ATTRIBUTES_FOR_TAGS
-        for value in topic[attribute]
-    ]
-
+    tags = []
+    for attribute in ATTRIBUTES_FOR_TAGS:
+        if attribute in topic and topic[attribute]:
```
Pierlou, line 82:
```suggestion
        if topic.get(attribute):
```

dgv/simplifions/delete_demo_topics.py
JeSuisUnCaillou: Yep, but i like versionning one shot scripts too, I might reuse it and I don't want to have it lost in some folder on my local machine.

JeSuisUnCaillou:
> Not tested so I'm trusting you on the process, only code suggestions (and an other one: we usually keep two blank lines before a function definition 🤓)

Yeah, I need to build some automated tests. I swear I'll do it at some point, 'cause this is already too boring to test by hand.

PR #547, opened 2025-08-12
Introduce mocked tests

JeSuisUnCaillou:


JeSuisUnCaillou:
Et cette PR merge vers cette autre PR sur laquelle j'avais commencé un taf avant de me rendre compte que sans tests j'allais probablement tout péter : https://github.com/datagouv/datagouvfr_data_pipelines/pull/546

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
```diff
@@ -0,0 +1,136 @@
...
+    "SIMPLIFIONS_produitspublics",
+    "SIMPLIFIONS_solutions_editeurs",
+    "SIMPLIFIONS_reco_solutions_cas_usages",
+    "Apidata",
+    "SIMPLIFIONS_description_apidata_cas_usages",
+]
+
+
+def request_grist_table(table_id: str, filter: str = None) -> list[dict]:
+    """Standalone function to request data from Grist table (copied from GristManager)."""
```
Pierlou, line 32: If it is exactly the same could we import it instead?

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
```diff
@@ -0,0 +1,136 @@
...
+            "table_id": table_id,
+            "total_records_available": len(raw_data),
+            "sample_record": sample_record,
+        }
+
+        # Save to current directory (we're already in the fixtures directory)
+        fixtures_dir = Path(__file__).parent
+
+        # Convert table name to filename (lowercase and replace underscores)
+        filename = table_id.lower().replace("_", "_") + ".json"
```
Pierlou, line 70: I am not sure I understand `.replace("_", "_")`

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def fetch_and_save_all_fixtures():
+    """Fetch data from all required Grist tables and save as fixtures."""
+
+    # Ensure we have the required environment variables
+    if not GRIST_API_URL or not SECRET_GRIST_API_KEY:
+        print("Error: GRIST_API_URL and SECRET_GRIST_API_KEY must be set")
+        print("Please set these environment variables before running this script.")
+        return False
```
Pierlou, line 102: You could also `raise` an Exception to avoid having to return True/False and catch the output further down

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # Save to current directory (we're already in the fixtures directory)
+        fixtures_dir = Path(__file__).parent
+
+        # Convert table name to filename (lowercase and replace underscores)
+        filename = table_id.lower().replace("_", "_") + ".json"
+        fixture_file = fixtures_dir / filename
+
+        with open(fixture_file, "w", encoding="utf-8") as f:
+            json.dump(fixture_data, f, indent=2, ensure_ascii=False)
```
Pierlou, line 74: Do you have a process to clean up these files when the tests are done?

verticales/simplifions/tests/grist_mock.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Create a callback function to handle filtering
+        def response_callback(request, context):
+            filter_param = request.qs.get("filter", [None])[0]
+            filtered_records = self._filter_records(records, filter_param)
+            response_data = self._create_grist_response_for_records(filtered_records)
+            return json.dumps(response_data)
+
+        url_pattern = (
+            f"https://grist.example.com/api/docs/c5pt7QVcKWWe/tables/{table_id}/records"
```
Pierlou, line 45: You have the doc_id hard coded in `fixtures/generate_fixtures_from_grist.py`, maybe you'd want to have it variabilized just in case

verticales/simplifions/tests/test_grist_manager.py
```diff
@@ -0,0 +1,399 @@
...
+from pathlib import Path
+import pytest
+import requests_mock
+
+# Add the parent directory to the path so we can import the module
+sys.path.append(str(Path(__file__).parent.parent))
+
+# The factory must be imported before the manager because it initializes the mocks
+from grist_factory import GristFactory
+from grist_manager import GristManager
```
Pierlou, line 11: I don't see the `grist_manager` file here, where does this come from?

verticales/simplifions/tests/grist_factory.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def create_records(
+        self, table_id: str, count: int, data: dict = None
+    ) -> list[dict]:
+        records = self.build_records(table_id, count, data)
+        self.records[table_id].extend(records)
+        grist_mock.mock_table_with_data(table_id, self.records[table_id])
+        return records
+
+    def create_record(self, table_id: str, data: dict = None) -> dict:
```
Pierlou, line 34: To be aligned with the `create_records` function, you could have this return `[records[0]]` and avoid postprocessing downstream

verticales/simplifions/tests/test_grist_manager.py
```diff
@@ -0,0 +1,399 @@
...
+        result = grist_manager._get_subdata("API_et_donnees_utiles", None, table_info)
+        assert result is None
+
+        # Test with empty string
+        result = grist_manager._get_subdata("API_et_donnees_utiles", "", table_info)
+        assert result == ""
+
+        # Test with empty list
+        result = grist_manager._get_subdata("API_et_donnees_utiles", [], table_info)
+        assert result == []
```
Pierlou, line 201: NIT: The cleanest way to do such tests is with `parametrize` (see [this](https://github.com/datagouv/csv-detective/blob/6dd1992f9fb88f58c88155638cf374c57795a76f/tests/test_fields.py#L391) for instance)

Pierlou (commented):
Wow this escalated quickly, huge work 👏

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
JeSuisUnCaillou: This script is a one shot to generate the fixtures files.
I'll eventually use it again to update them if we modify the structure in grist, so I want to keep it. But it's not run during the tests at all.

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
JeSuisUnCaillou: At first I agreed. Then I started to import the method and I remembered :

That's because in the grist_manager I use env vars from airflow, and here I use local env vars (since this is a one shot script I only use to generate the fixtures locally, cf https://github.com/datagouv/datagouvfr_data_pipelines/pull/547#discussion_r2272398480)

I'll refacto a bit the method to have a common method I can use with these vars in parameters.

verticales/simplifions/tests/fixtures/generate_fixtures_from_grist.py
JeSuisUnCaillou: Lol, yeah that's stupid. I'm removing that, thanks :D

verticales/simplifions/tests/grist_mock.py
JeSuisUnCaillou: This doc_id has nothing to do with `fixtures/generate_fixtures_from_grist.py`, this script is not used during the tests runs. It is related to the GRIST_DOC_ID in the grist_manager, though. 

But I think it would be better to catch all patterns with any doc_id in this grist_mock. I'll do that.

PR #570, opened 2025-09-22
dgv.metrics: fix duplicated aggregations when task retry

hacherix:


maudetes (approved):
:pray: thank you for the fix

PR #574, opened 2025-09-30
add DAG verticale_simplifions_grist_watcher

JeSuisUnCaillou:


verticales/simplifions/grist_v2_manager.py
```diff
@@ -42,6 +45,30 @@ def _request_grist_table(self, table_id: str, filter: str = None) -> list[dict]:
         r.raise_for_status()
         return r.json()["records"]
 
+    def _request_all_tables(self) -> dict:
```
Pierlou, line 48: These functions (and the one above too) should be static methods I think, they belong in the class but don't use the class instance in their process

verticales/simplifions/grist_v2_manager.py
```diff
@@ -88,3 +115,78 @@ def get_and_format_grist_v2_data(self, ti):
...
+
+                if modified_rows and len(modified_rows) > 0:
+                    logging.info(
+                        f"> Found {len(modified_rows)} modified rows since {time_delta} hours ago"
+                    )
+                    tables_with_modified_rows[table["id"]] = modified_rows
+                else:
+                    logging.info(f"> No modified rows found for table {table['id']}")
+            else:
+                logging.info("> Does not have a 'Modifie_le' column. Skipping.")
```
Pierlou, line 156: NIT: a slightly better syntax for these cases could be:
```python
if "Modifie_le" not in [column["id"] for column in table_columns]:
    logging.info("> Does not have a 'Modifie_le' column. Skipping.")
    continue
logging.info("\n")
logging.info(f"Table: {table['id']}")
...
```

verticales/simplifions/grist_v2_manager.py
```diff
@@ -88,3 +115,78 @@ def get_and_format_grist_v2_data(self, ti):
...
+                logging.info(f"> Found {len(all_rows)} rows total")
+
+                modified_rows = [
+                    row
+                    for row in all_rows
+                    if row["fields"].get("Modifie_le")
+                    and row["fields"]["Modifie_le"] > cutoff_timestamp
+                ]
+
+                if modified_rows and len(modified_rows) > 0:
```
Pierlou, line 148: `if my_list` is the same as `if len(my_list) > 0` so it's redundant
```suggestion
                if modified_rows:
```

verticales/simplifions/grist_v2_manager.py
```diff
@@ -88,3 +115,78 @@ def get_and_format_grist_v2_data(self, ti):
...
+            else:
+                logging.info("> Does not have a 'Modifie_le' column. Skipping.")
+
+        # Format the message
+
+        tables_messages = []
+        message = "# Modifications du grist Simplifions" + "\n"
+        message += f"Les données suivantes ont reçu des modifications pendant les {time_delta} dernières heures:\n"
+
+        message += "\n\n"
```
Pierlou, line 164:
```suggestion
        message += f"Les données suivantes ont reçu des modifications pendant les {time_delta} dernières heures:\n\n\n"
```

verticales/simplifions/grist_v2_manager.py
```diff
@@ -88,3 +115,78 @@ def get_and_format_grist_v2_data(self, ti):
...
+                    tables_with_modified_rows[table["id"]] = modified_rows
+                else:
+                    logging.info(f"> No modified rows found for table {table['id']}")
+            else:
+                logging.info("> Does not have a 'Modifie_le' column. Skipping.")
+
+        # Format the message
+
+        tables_messages = []
+        message = "# Modifications du grist Simplifions" + "\n"
```
Pierlou, line 161:
```suggestion
        message = "# Modifications du grist Simplifions\n"
```

verticales/simplifions/grist_v2_manager.py
```diff
@@ -88,3 +115,78 @@ def get_and_format_grist_v2_data(self, ti):
...
+            table_message = f"### **{table_id}**\n"
+
+            for modifier, rows in modified_rows_grouped_by_modifier.items():
+                table_message += f"- **{modifier}**\n"
+                for row in rows:
+                    table_message += f"  - {row['fields']['technical_title']}\n"
+
+            table_message += "\n"
+
+            tables_messages.append(table_message)
```
Pierlou, line 182:
```suggestion
            tables_messages.append(table_message + "\n")
```

verticales/simplifions/task_functions.py
```diff
@@ -12,3 +12,7 @@ def get_and_format_grist_v2_data(ti, client=None):
 
 def update_topics_v2(ti, client=None):
     return TopicsV2Manager(client).update_topics(ti)
+
+
+def watch_grist_data(ti):
```
Pierlou, line 17: As a general note, I find it twisted to have the task functions in the classes and to call them with one-liners like this; I would rather keep only the "utils" functions in the classes (`_request_grist_table` etc) and move the task functions in here

Pierlou (commented):
The flow is good 👏 only refactoring suggestions for what I believe would be better syntaxes

verticales/simplifions/task_functions.py
JeSuisUnCaillou: It's gonna require some annoying refacto, but yes you're right.

PR #575, opened 2025-10-01
[DAG HVD] refactor: prepare for dashboard

Pierlou:


dgv/monitoring/hvd/task_functions.py
```diff
@@ -428,6 +429,20 @@ def build_df_for_grist():
 
 
 def update_grist(ti):
+    def update_quality(dataset_id: str):
+        r = requests.get(
+            f"https://www.data.gouv.fr/api/1//datasets/{dataset_id}/",
```
maudetes, line 434: Could we use api 2 in this case?
```suggestion
            f"https://www.data.gouv.fr/api/1/datasets/{dataset_id}/",

```

dgv/monitoring/hvd/task_functions.py
```diff
@@ -428,6 +429,20 @@ def build_df_for_grist():
...
+        r = requests.get(
+            f"https://www.data.gouv.fr/api/1//datasets/{dataset_id}/",
+            headers={"X-fields": "quality"},
+        )
+        if not r.ok:
+            logging.error(f"Issue with {dataset_id}")
+            return
+        dataset_quality = r.json()["quality"]
+        table.update_records(
+            conditions={"id2": dataset_id},
```
maudetes, line 442: why the sad `id2` name? Could it be `dataset_id` :stuck_out_tongue:  ?

dgv/monitoring/hvd/task_functions.py
```diff
@@ -493,39 +508,58 @@ def update_grist(ti):
...
+                new_values={"missing_hvd_tag": True},
+            )
+        send_message(message, MATTERMOST_MODERATION_NOUVEAUTES)
+
+    # updating quality metrics
+    logging.info("Updating datasets quality metrics columns")
+    df = table.to_dataframe(columns_labels=False, usecols=["id2"])
+    df["id2"].apply(update_quality)
+    file_name = "grist_hvd.csv"
+    table.to_dataframe().to_csv(DATADIR + file_name, sep=";", index=False)
```
maudetes, line 550: shouldn't we use `df.to_csv()`? Is this going to be used for monitoring? and then don't we want to update a file on minio instead of having one file by date?

maudetes: Yep, I thought that update_quality may update the df itself but nope.

I was wondering for a unique file with month information inside, but let's keep it as it is :pray:

maudetes (approved):

PR #577, opened 2025-10-03
[DGV metric] add indexes on commonly used tables

maudetes:
The datagouv frontend calls for metrics when visiting an object page.
Exemple on https://www.data.gouv.fr/datasets/paris-2024-sites-de-competition/, we have the two calls :
* https://metric-api.data.gouv.fr/api/datasets/data/?dataset_id__exact=65df34eeb2a9d54adad981d7&metric_month__sort=desc&page_size=12 
* https://metric-api.data.gouv.fr/api/datasets_total/data/?dataset_id__exact=65df34eeb2a9d54adad981d7

They take up to 4sec, which displays a 0 on the metrics graphs before the api calls return.
image

I've added the same indexes for all objects, since I wanted to be consistent, but it may be overkill for some.

### DIFF
--- a/dgv/metrics/sql/create_tables.sql
+++ b/dgv/metrics/sql/create_tables.sql
@@ -337,6 +337,7 @@ CREATE MATERIALIZED VIEW IF NOT EXISTS metric.dataservices_total AS
     GROUP BY dataservice_id
 ;
 
+-- Index on visits table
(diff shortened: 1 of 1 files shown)


hacherix (approved):
Niice! 🐹

PR #615, opened 2026-01-08
simplifions: change schedule for verticale_simplifions_v2_demo

abulte:
Do not run at night, since demo.data.gouv.fr is more susceptible to fails.


Pierlou (approved):
🕥

PR #629, opened 2026-02-13
feat(simplifions): add categorie-de-solutions tags

abulte:
Related https://linear.app/pole-api/issue/API-6339/ajouter-le-nouveau-filtre-categorie-de-solution-et-supprimer-le-filtre


Pierlou (approved):
Trusting you on this!

PR #653, opened 2026-04-16
Migrate DVF geoloc to Airflow

Pierlou:
Transcription of https://github.com/datagouv/dvf


data_processing/dvf/explore/dag.py
```diff
@@ -92,8 +92,7 @@
         )
 
     (
-        clean_up_folder(TMP_FOLDER, recreate=True)
```
estellebertrand, line 95: So now cleanup is at the end and not at the beginning right ?

data_processing/dvf/geoloc/task_functions.py
```diff
@@ -0,0 +1,303 @@
...
+
+@task()
+def download_source_data(**context):
+    dvf_dataset = Dataset(SOURCE_DATASET_ID)
+    data = [res for res in dvf_dataset.resources if res.type == "main"]
+    if len(data) not in {5, 6}:
+        # 5 full years, or half first and last years and 4 full years
+        raise ValueError(f"Unexpected number of resources: {len(data)}")
+    files = []
+    max_year = 2000
```
estellebertrand, line 49: Unsure here what does the max year represent ? is it 2000 or 2020 ? (as i don't find a file related to 2000 but i don't know much about this dataset)

estellebertrand: since there is "    # one major version per year since 2020", i guess max_year must be above 2020 no?

data_processing/dvf/explore/dag.py
Pierlou: The data folder is shared between the two DAGs, they run one after the other and we keep the geoloc files at the end of geoloc to process them in explore. That prevents a re-download

data_processing/dvf/geoloc/task_functions.py
Pierlou: These years reference the max year of the DVF data (10 rolling semesters)
The comment is about the cadastre versions

data_processing/dvf/explore/scripts/script_dl_geo.sh
```diff
@@ -16,12 +16,4 @@ curl https://cadastre.data.gouv.fr/data/dgfip-pci-vecteur-latest.json > $DATADIR
 
 jq -r '.[].contents[] | select(.name | endswith("edigeo")) | .contents[] | select(.name | endswith("feuilles"))| .contents[].contents[].contents[].name' $DATADIR/dgfip-pci-vecteur-latest.json | cut -d '/' -f 11 | sed 's/edigeo-\|\.tar\.bz2//g' >| $DATADIR/sections.txt
 
-for YEAR in `seq $five_ago $curr_year`
```
estellebertrand, line 19: To understand, why are we deleting that ? done somewhere else?

data_processing/dvf/geoloc/task_functions.py
estellebertrand: Ok can you confirm that is correct: we look at  resources`res for res in dvf_dataset.resources` from dataset `5c4ae55a634f4117716d5656`, all of which are above 2020 in their title pattern, then we take the max_year between 2000 and 2021+, so it ends up being 2025, then we do `version = max_year - 2020` so we ends up with version 5 for the cadastre ? (currently)

data_processing/dvf/explore/scripts/script_dl_geo.sh
Pierlou: Yes, we will already have the file on the machine, from the previous DAG (that creates them)

data_processing/dvf/geoloc/task_functions.py
Pierlou: Yes exactly, 2020 is the year of the first release (0.x.y), every year sees a major version release, so we fetch the file for the latest year.
However currently the output of this part of the code (`arrondissements_muni`) is not used afterwards (see the following task `enrich_years`), so it doesn't really matter for now, and could change in the future

data_processing/dvf/geoloc/task_functions.py
```diff
@@ -257,7 +271,7 @@ def enrich_year(
         ]
         geoloced.append(enriched.dropna(subset="longitude"))
     geoloced = pd.concat(geoloced + [remainders], ignore_index=True).sort_values(
-        by="id_mutation", key=lambda col: col.str.split("-").str[1].astype(int)
+        by="id_mutation", key=lambda col: col.str.split("-").str[1].astype(int), axis=1
```
estellebertrand: I'm confused, why `axis=1` ? We want to sort on the index axis by id_mutation no ?

data_processing/dvf/geoloc/task_functions.py
```diff
@@ -0,0 +1,374 @@
...
+    output["surface_terrain"] = (
+        source["Surface terrain"].str.replace(",", ".").astype(float)
+    )
+    del source
+
+    logging.info("Creating mutation ids...")
+    # sorting to group mutations in the dataframe
+    output.sort_values(by=["date_mutation", "valeur_fonciere"], inplace=True)
+    # new mutation id when either date or price changes
+    mask = (output["date_mutation"] != output["date_mutation"].shift()) | (
```
estellebertrand, line 240: if both values for one of the col (date or price) equal NaN then it will be marked True in the mask, is it expected (for instance, impossible for these col to be NaN ever) ?

data_processing/dvf/geoloc/task_functions.py
Pierlou: I mixed up with `apply` and `rename`, I'll revert that, thanks for spotting

data_processing/dvf/geoloc/task_functions.py
Pierlou: I checked for the 2025 file: date is never None, but price can be, so all transactions that occured on the same date but with no disclosed price will be attributed the same id. Maybe we change that so that such mutations would all have different ids, or we use an other column to group them (parcelle_id for instance)
We should keep in mind that these ids are attributed in this process but have no official value, and are often not even stable from one delivery to another

estellebertrand: It doesn't seem possible but technically could a parcelle be sold twice the same day ? ^^ For the mask we can also just replace nan value by negative value for instance like -1 (if that's well a value never possible for a price) like `output["valeur_fonciere"].fillna(-1) != output["valeur_fonciere"].fillna(-1).shift()` to get different ids 
Even though no official value, it might be used by some users to dedupe I guess?

Pierlou: > Maybe we change that so that such mutations would all have different ids

Actually it's already the case, NaN != NaN is true, so such mutations have unique ids, that looks fine to me

estellebertrand (approved):
Looks good ! Big work !

PR #654, opened 2026-04-20
[petitions] Fix max id retrieval

Pierlou:
From a smart idea suggested in https://github.com/datagouv/datagouvfr_data_pipelines/pull/652


estellebertrand (approved):
It looks ok, but cannot test it fully locally because it requires to download them all !

PR #657, opened 2026-05-13
fix: default tchap room

Pierlou:


utils/tchap.py
```diff
@@ -34,6 +34,11 @@ def send_message(
     ping: list[str] = [],
 ) -> None:
     """Send a message to a Tchap channel."""
+    if not room_id:
+        if AIRFLOW_ENV == "prod":
+            raise ValueError("The room id is not set")
+        else:
+            room_id = TCHAP_ROOM_DATAENG_TEST
```
estellebertrand, line 43: We also need to have the check on TCHAP_BASE_URL, because the error is not explicit when this var is not set. Same for the token i'd say. I believe it might be change to an HTTP connexion to make it cleaner with host and secret maybe?

Pierlou: Done in https://github.com/datagouv/datagouvfr_data_pipelines/pull/657/commits/fbf43fd07e3f79355bb8172d590fdb46974b9059, we can move to a connection in a second PR (needs to be set up through `infra`)

PR #658, opened 2026-05-13
chore: correcting meta and maintenance DAGs for Airflow3

estellebertrand:
This PR encompasses : 

**1. A new `utils` module :** an Airflow API connector, implementing the new correct way to communicate with Airflow DB in v3

**2. An upgrade to the DAG `maintenance_clean_logs_and_runs` :** now uses the airflow API to access DagRun objects

**3. Multiples upgrades to the DAG `meta_dag`:** 
  - now uses the airflow API to access Dag, DagRun and TaskInstance objects
  - wait until the end of the DAG to ~~raise an error~~ send a Tchap notification in case of DAGs name not being matched in the DB : both 1. to help with maintenance of the file `meta/config.json` if not matching the DB anymore, and 2. for testing (it allows the DAG to run in local when some DAGs are missing - this is expected due to environment diff between dev and prod)
  ~~- added to the DAGs to monitor in `meta/config.json` so the next successful run can detect an issue with the prior, in case no one see the notifications are missing~~

**Note: this PR requires apache-airflow-client to be added to the infra to run, it will throw an error otherwise.**

### DIFF
--- a/maintenance/dag_clean_logs_and_runs.py
+++ b/maintenance/dag_clean_logs_and_runs.py
@@ -5,13 +5,14 @@ from datetime import datetime, timedelta, timezone
 
 from airflow import DAG
 from airflow.decorators import task
-from airflow.models import DagRun
-from airflow.settings import Session
+from airflow_client.client.api import dag_run_api
+
 from datagouvfr_data_pipelines.config import (
     AIRFLOW_DAG_HOME,
     AIRFLOW_ENV,
 )
 from datagouvfr_data_pipelines.utils.tchap import send_message
+from datagouvfr_data_pipelines.utils.airflow import AirflowAPI
 
 nb_days_to_keep = 60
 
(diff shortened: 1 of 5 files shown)


maintenance/dag_clean_logs_and_runs.py
Pierlou, line 1: For the logs the current process still works?

meta/config.json
```diff
@@ -36,8 +36,12 @@
     "schema_consolidation": true,
     "irve_consolidation": true,
     "maintenance_clean_logs_and_runs": true,
+    "meta_dag": true,
```
Pierlou: Isn't that useless? If the meta DAG fails, we won't have the notif (as it failed), and if we get the notif, it means it ran successfully?
Update: after seeing the newly added task I understand, but maybe a ping is better than a fail + report
```suggestion
```

meta/task_functions.py
```diff
@@ -24,68 +28,85 @@
...
+                    dag_id=dag_id,
+                    end_date_gte=date_dt,  # Does not filter out currently running DAGs
+                ).dag_runs
+            except NotFoundException:
+                dags_not_found.append(dag_id)
+                continue  # Raise an error only at the end to avoid skipping the rest of DAGs
+            for dag_run in dag_runs:
+                start_date, end_date = dag_run.start_date, dag_run.end_date
+                if not start_date:
+                    raise TypeError(f"Start date must be defined for DagRun {dag_run}")
```
Pierlou: In which case does this happen?

meta/task_functions.py
```diff
@@ -168,3 +189,14 @@ def notification(**context):
...
     send_message(message, ping=list(ping))
+
+
+@task()
+def raise_dags_not_found(**context):
+    dags_not_found = context["ti"].xcom_pull(
+        key="dags_not_found", task_ids="monitor_dags"
+    )
+    if len(dags_not_found) > 0:
+        raise NotFoundException(
```
Pierlou: Could we have a Tchap ping instead of an error here?

Pierlou (commented):
Thanks for fixing these two DAGs 🙏

maintenance/dag_clean_logs_and_runs.py
estellebertrand: You mean the code before this PR ? Not on my machine !

meta/task_functions.py
estellebertrand: More of a legacy when I tested start_date/end_date, as we compute end_date-start_date later. But indeed it should logically not happen, I can remove it

meta/task_functions.py
estellebertrand: Yes I'll add that, good idea ! Even though, I find more logical to keep this error (even if we add a Tchap ping before), as technically it is an error to correct if some DAGs are not found. What do you think?

meta/task_functions.py
estellebertrand: PS: I remember now, it throws a Pylance error 
```
Operator "-" not supported for types "datetime" and "datetime | None"
Operator "-" not supported for types "datetime" and "None"
```
but I'll ignore it

meta/task_functions.py
estellebertrand: I removed the error and send a tchap notif

maintenance/dag_clean_logs_and_runs.py
Pierlou: Weird, do you have all the required variables properly set up? Idk how the logs are handled in Airflow 3, maybe the current process still works, but maybe not 😕

meta/task_functions.py
Pierlou: I see, indeed IRL it should never happen (or I guess we'd be happy to have a crash to debug if it ever does)

Pierlou (approved):
Thanks 🙏  I'm just wondering about the logs deletion process, maybe we can ask ADE what they do
Update: ADE says it still works, so here we go!

PR #659, opened 2026-05-18
Add dags decoupage_administratif and contours_administratifs

ThomasG77:
We need to update resources published on https://www.data.gouv.fr/datasets/decoupage-administratif-1 and https://www.data.gouv.fr/datasets/contours-administratifs annually but the process was not documented, only in a local script for a first run.
These resources are used for running the API Découpage Administratif
This PR purpose is to solve the issue.

### DIFF
--- a/data_processing/contours_administratifs/task_functions.py
+++ b/data_processing/contours_administratifs/task_functions.py
@@ -1,0 +1,185 @@
+import json
+import logging
+import mimetypes
+import re
+
+
+from airflow.decorators import task
+
+# from datagouv import Client
+from datagouvfr_data_pipelines.config import (
+    AIRFLOW_ENV,
+    AIRFLOW_DAG_HOME,
+    S3_URL_RBX,
+)
+from datagouvfr_data_pipelines.utils.datagouv import (
+    local_client,
+)
+from datagouvfr_data_pipelines.utils.s3 import S3Client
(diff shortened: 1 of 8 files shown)


data_processing/contours_administratifs/task_functions.py
```diff
@@ -0,0 +1,185 @@
...
+from datagouvfr_data_pipelines.utils.datagouv import (
+    local_client,
+)
+from datagouvfr_data_pipelines.utils.s3 import S3Client
+from datagouvfr_data_pipelines.utils.tchap import send_message
+import requests
+
+mimetypes.init()
+
+# Variables directly below need an update each year
```
estellebertrand, line 24: Do they change each year in a way that makes it impossible to update it programmatically ? Like a certain date you can't know in advance

data_processing/decoupage_administratif/config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "decoupage-administratif": {
+        "dev": {
+            "dataset_id": "683424208089f77caf78f448"
```
estellebertrand, line 4: is it normal that they have the same ID on demo and prod?

estellebertrand (approved):
It looks ok, but I don't know the original script so can't compare really ! Has it been tested locally ? If yes seems good. Just a few comments :)

PR #661, opened 2026-05-22
Use Airflow SDK to remove deprecaction warnings and lint

Pierlou:


estellebertrand (approved):
I trust the syntax is good, looks good :) merge conflict to fix first, markdown import can be removed (already below) but logging must stay

PR #662, opened 2026-05-26
fix: add pagination to AirflowAPI and state filter

estellebertrand:
**Issues:**
1. It wasn't shared explicitly in the apache-airflow-client that the client itself also requires pagination (like the API)
2. The client to get DagRuns doesn't filter out runs without end_date when filtering on it (like `end_date_lte`)

**Solutions:**
1. A pagination method has been added to AirflowAPI so it can be shared across DAGs using it
2. A filter on `state` has been added to ensure no running nor queued runs are returned

Both fix have been made to Meta and Maintenance DAGs.

### DIFF
--- a/utils/airflow.py
+++ b/utils/airflow.py
@@ -58,3 +58,26 @@ class AirflowAPI:
 
         # Instance of the API client to use as a context
         self.client = airflow_client.client.ApiClient(configuration)
+
+    @staticmethod
+    def paginate(api_call, result_attr: str, page_size: int = 100, **kwargs):
+        """Paginate through all results of an Airflow API list call.
+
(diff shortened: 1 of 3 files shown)


utils/airflow.py
```diff
@@ -58,3 +58,26 @@ def __init__(self, conn_name):
 
         # Instance of the API client to use as a context
         self.client = airflow_client.client.ApiClient(configuration)
+
+    @staticmethod
+    def paginate(api_call, result_attr: str, page_size: int = 100, **kwargs):
```
Pierlou, line 63: It took me a while to understand the `api_call` object, could you please a hint type (`typing.Callable` should be fine) 🙏

Pierlou (approved):
Nice, thanks for the fix 🤞

PR #672, opened 2026-06-19
Periodically remove old digests

Pierlou:


maudetes (commented):
I think I would destroy the daily ones but keep the weekly as archive?
The annual digest is not equal to the sum of the monthly ones since there are deletions/archives

Pierlou:
> I think I would destroy the daily ones but keep the weekly as archive? The annual digest is not equal to the sum of the monthly ones since there are deletions/archives

Fair enough, that's most of the mass, done in https://github.com/datagouv/datagouvfr_data_pipelines/pull/672/commits/f68017f0359b258ece857e41803d47989476cf23

maudetes (approved):
Thank you!

PR #678, opened 2026-06-30
feat(geozones): rebuild parents/ancestors hierarchy from INSEE relations

ThibaudDauce:
### What & why
The geozones export now ships a full administrative hierarchy for every zone, and fixes several long-standing data defects surfaced while validating the export against the live INSEE source.

### New features
- **`parents` + `ancestors` on every record** — rebuilt from INSEE `subdivisionDirecteDe` relations. Parents climb over zones excluded from the export (e.g. départemental arrondissements), so a commune still exposes its department + EPCI; `ancestors` lists the full transitive chain up to `country:fr`.

### Bug fixes
- **Missing `nomSansArticle` / `codeArticle`** — the grouped `OPTIONAL` block only bound when *all* attributes matched, silently dropping these fields and mislabelling some deletions. Now split into independent `OPTIONAL`s.
- **Namibia had no id** — its ISO code `NA` was parsed by pandas as `NaN`, leaving `_id`/`codeINSEE` null. Read the countries CSV with `keep_default_na=False` → valid `country:na`.
- **Robustness** — `raise_for_status()` on INSEE calls, plus guards that abort if the hierarchy query returns no rows / unexpected columns, and a null-`_id` guard.

### Impact on the exported file (validated against prod 2024)
|                 | Prod 2024 | This PR (2026)                |
| --------------- | --------- | ----------------------------- |
| Records         | 53,752    | 54,780                        |
| Keys per record | 9         | 11 (`+parents`, `+ancestors`) |

**2026 data refresh** (vs prod): **+667** new ids (620 communes, 45 departments/overseas, 2 EPCI), **0** zones lost · 149 renames · 186 deletion-status changes · 99 type changes.

**Specific corrections** (on records common to both):
- `nomSansArticle`: **39,014 null → 0** (~36k names recovered)
- `codeArticle`: ~35,900 recovered
- Records with null `_id`: **1 → 0** (Namibia)
- `parents`/`ancestors` populated for all active zones; deleted zones correctly have no parents.

### DIFF
--- a/data_processing/geozones/task_functions.py
+++ b/data_processing/geozones/task_functions.py
@@ -1,5 +1,7 @@
 import json
+import logging
 import os
+from collections import defaultdict
 from datetime import datetime
 from io import BytesIO
 from urllib.parse import quote_plus, urlencode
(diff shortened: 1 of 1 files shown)


estellebertrand (approved):
Let's see if that works well :)

PR #681, opened 2026-06-30
Alert if non unique file id in dataset

Pierlou:
Raises an alert if there are more than one file with the same id in a dataset (for instance [here](https://www.data.gouv.fr/datasets/donnees-changement-climatique-sim-quotidienne))


estellebertrand (approved):
Ok looks good thanks

PR #682, opened 2026-07-01
feat: add population and geometries

ThibaudDauce:
Enrich every geozone with two new fields, joined on the INSEE code:

### `geom` — administrative contour
- Source: the **contours-administratifs** bucket (IGN ADMIN EXPRESS, what `geo.api.gouv.fr` is built on) at **100 m** resolution. INSEE SPARQL serves no geometry.
- Communes merge the associated/delegated file, so merged communes keep a contour. **~99 %** of active zones get a geometry; countries and municipal arrondissements (no published source) keep `null`.

### `population` — INSEE legal population
- Source: **geo.api.gouv.fr** (recent census millésime). Communes and EPCI come directly; department and region populations are **summed from their communes** (geo.api doesn't serve them). Validated: active regions sum to 68.35 M.
- **~94 %** of active zones covered; merged communes / countries keep `null`.
- INSEE SPARQL also exposes population (`demo:` vocabulary) but is **frozen at the 2016 millésime**, so geo.api is used instead.

### DIFF
--- a/data_processing/geozones/task_functions.py
+++ b/data_processing/geozones/task_functions.py
@@ -230,6 +230,56 @@ def fetch_geozones_geometries() -> dict:
     return geometries
 
 
+POPULATION_ENDPOINT = "https://geo.api.gouv.fr"
+
+
+def fetch_geozones_populations() -> dict:
+    """
+    Return the legal population per zone as {level: {code: population}}.
+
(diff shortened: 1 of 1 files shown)


estellebertrand (approved):
okkk

PR #684, opened 2026-07-02
fix(geozones): upload datasets and import magic timeout

ThibaudDauce:
not sure about the `import magic` timeout, it doesn't seems to be related to my changes. But it's cleaner to import it inside the function if it's slow to import…


estellebertrand (approved):
Hope it solves it!

PR #687, opened 2026-07-03
ci: add pre-commit

hacherix:
Ajout de la config pre-commit inspirée de https://github.com/annuaire-entreprises-data-gouv-fr/search-infra/blob/main/.pre-commit-config.yaml
À adapter en fonction de notre usage au fur et à mesure.


estellebertrand (commented):
Has it been tested yet ? does this config change anything in the repo (like current styling) ?

hacherix:
> Has it been tested yet ?

Yep it works well. The same config is used on the Annuaire Airflow and has been battle tested.

> does this config change anything in the repo (like current styling) ?

Yep the ruff `I` rule and the `pre-commit-hooks` do change a few things. Would you prefer if we fix all the files now or we let them be fixed organically whenever we modify a file in a PR?

`mypy` is raising quite a lot of issues with the typing though. Same, I can either fix them all now or not.

Both solutions are fine for me. What do you prefer? :smile:

estellebertrand:
I'd say let's fix them all in one go to avoid mental charge when reviewing future PR ^^

hacherix:
> I'd say let's fix them all in one go to avoid mental charge when reviewing future PR ^^

Done ! Mostly typing and variable name potential conflicts issues.
But there is one bug found by mypy fixed in the last commit.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: 'v0.15.20'  # Ruff version
+    hooks:
+      - id: ruff-check
+        args: [--fix, --extend-select, I]
+      - id: ruff-format
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v2.1.0'
+    hooks:
+      - id: mypy
```
bolinocroustibat, line 22: A nicer (faster and stricter) option in the future could be to use https://github.com/allganize/ty-pre-commit

hacherix: Carrément ! Faudrait passer le repo sur `uv` d'abord pour vraiment bénéficier de `ty`. Il est plus complexe à setup que `mypy` je trouve.
Il y a un repo officiel aussi : https://github.com/astral-sh/ty-pre-commit

bolinocroustibat: > Carrément ! Faudrait passer le repo sur `uv` d'abord pour vraiment bénéficier de `ty`. Il est plus complexe à setup que `mypy` je trouve. Il y a un repo officiel aussi : https://github.com/astral-sh/ty-pre-commit

Ah oui beaucoup mieux le repo officiel effectivement, je ne connaissais pas.
Aucune urgence, c'était une petite pensée en passant par là, vraiment pas prioritaire.

estellebertrand (approved):

PR #694, opened 2026-07-06
fix: avoid redundant s3 client instanciation

estellebertrand:
## Problem

S3 client was instanciated for each file, knowing that a single loop running late could be as much as 130K files, it is a lot of unnecessary instanciations and could add charge to RAM usage (likely not the cause of OOM errors in itself but could add pressure).

## Proposal

Instanciate S3Client just once as it can be shared across all files being processed, `s3_client_kwargs` being a constant.


hacherix (approved):
Nice one!

PR #700, opened 2026-07-07
fix(meteo): handle missing TMP_FOLDER in clean_directory

bolinocroustibat:
## Problem
`clean_directory` called `os.listdir`on `TMP_FOLDER` before `create_working_dir` had created it, so the task was crashing on first run.

## Fix
The fix checks that the directory exists and returns early if it doesn’t, using `pathlib` for the cleanup loop.


estellebertrand (approved):
Good we should do the same on all repo requiring folder cleaning

PR #704, opened 2026-07-09
fix(meteo): notifications fix, add log and reuse variable

maudetes:
- fix notification task with iteration on `.items()` (see [error](https://workflows.infra.data.gouv.fr/dags/data_processing_meteo/runs/scheduled__2026-07-09T05:30:00+00:00/tasks/notification))
- add log before S3 re-fetch
- reuse `bucket` var


estellebertrand (approved):
Good catch ! Thank you

PR #725, opened 2026-07-29
chore: s3 connexions

estellebertrand:
## What this PR does

- update the S3 client so it uses a native Airflow connection instead of environment variables 
- add methods to the S3 client that are necessary for the integration of Datalma DAGs
- update all uses of the S3 client so the right connection is used (one per region: the region could also be used as a separate params therefore using a single shared connection for all, but that might not workout with the OVH url syntax, it needs to be investigated for a possible future PR)

## DO NOT MERGE AND PUSH

Before : 
- the connections have been created on the prod infra : <del>`S3_RBX`, `S3_SBG`, `S3_EU_WEST_PAR`</del> `S3_OVH_RBX`, `S3_OVH_SBG`, `S3_OVH_EU_WEST_PAR`.
- the required packages have been added to the infra* : `mypy_boto3_s3==1.42.67`

(*can add required package for Datalma as well at the same time :
`astronomer-cosmos==1.15.0`
`dbt-core==1.11.7`
`dbt-postgres==1.10.0`
`sqlalchemy==2.0.48`)

## Warning

This PR cannot be tested on dev because the required buckets do not exist in a dev env and no safe DAG has been identified for testing in prod locally. Therefore once merged and pushed, this might affect all DAGs using S3, including critical ones. A careful review is necessary.


verticales/simplifions/tests/conftest.py
```diff
@@ -55,6 +55,7 @@ def decorator(f):
 config_mock.GRIST_API_URL = "https://grist.example.com/api/"
 config_mock.SECRET_GRIST_API_KEY = "test-api-key"
 # S3 config
+# TODO : find out how those vars are used and it there is an issue with the new S3Client
 config_mock.S3_URL_EU_WEST = "s3.example.com"
```
bolinocroustibat, line 59: Those are the old vars no? Shouldn't we remove them?

utils/s3.py
```diff
@@ -1,42 +1,49 @@
...
-)
+from mypy_boto3_s3.service_resource import S3ServiceResource
+import requests
+
+from datagouvfr_data_pipelines.config import AIRFLOW_DAG_TMP, AIRFLOW_ENV
 from datagouvfr_data_pipelines.utils.filesystem import File
 from datagouvfr_data_pipelines.utils.retry import simple_connection_retry
 
 
+class S3ClientKwargs(TypedDict, total=False):
```
bolinocroustibat, line 19: Isn't a dedicated dataclass a bit overkill for this?
We could just do `s3_client = S3Client(bucket=S3_BUCKET_PNT, conn_name=S3_CONN_RBX)`, might be more readable and easier to maintain. What do you think?

data_processing/meteo/previsions_densemble/task_functions.py
```diff
@@ -236,19 +231,16 @@ def fix_title(file_name: str) -> str:
...
+    for obj, size in s3_client.get_all_files_names_and_sizes_from_parent_folder(
+        folder=f"{AIRFLOW_ENV}/{s3_folder}/{pack}/{grid}/",
+    ).items():
         try:
             file_id, file_date = build_file_id_and_date(obj.split("/")[-1])
             if file_id not in latest_files or file_date > latest_files[file_id]["date"]:
                 latest_files[file_id] = {
                     "date": file_date,
-                    "url": f"https://{bucket_pe}.{S3_URL_RBX}/{obj}",
+                    "url": f"https://{bucket_pe}.{s3_client.url}/{obj}",
```
bolinocroustibat: We could use the shared function `get_file_url` from s3.py here for easier maintenance, no?

data_processing/meteo/previsions_numeriques_temps/task_functions.py
```diff
@@ -340,7 +335,7 @@ def publish_on_datagouv(model: str, pack: str, grid: str, **kwargs):
             if file_id not in latest_files or file_date > latest_files[file_id]["date"]:
                 latest_files[file_id] = {
                     "date": file_date,
-                    "url": f"https://{S3_BUCKET_PNT}.{S3_URL_RBX}/{obj}",
+                    "url": f"https://{S3_BUCKET_PNT}.{s3_pnt.url}/{obj}",
```
bolinocroustibat: Same here, we could use the shared function `get_file_url` from `s3.py` here maybe?

verticales/simplifions/tests/conftest.py
bolinocroustibat: Ha, I guess tests will still use those old vars?
Sorry not very familiar with this repo!

Maybe we could mock Airflow connection in tests instead?

utils/s3.py
```diff
@@ -1,42 +1,49 @@
...
+            S3ServiceResource,
+            boto3.resource(
+                "s3",
+                endpoint_url=f"https://{self.url}",
+                aws_access_key_id=self.login if login else None,
+                aws_secret_access_key=self.password if login else None,
+                config=Config(**config_kwargs),
+            ),
         )
         if bucket not in [b.name for b in self.resource.buckets.all()]:
```
bolinocroustibat, line 48: We might need to remember that `S3Client` always list buckets when starting, which might be costly and might fail if the user doesn't have the right to list buckets.
I don't know the repo very well but maybe we could think about doing it differently in a future PR?

bolinocroustibat (commented):
Thank you!!

Old naming `S3_URL_*` was not a good one, it was not a URL but a part of an URL which was a bit misleading and error-prone, good think we're getting rid of it!

A general remark: PR description says prod connexions are S3_RBX, S3_SBG, S3_EU_WEST_PAR, but the code uses S3_OVH_RBX, S3_OVH_SBG, S3_OVH_EU_WEST_PAR. Which one is right?

data_processing/meteo/previsions_densemble/task_functions.py
estellebertrand: Yes indeed, i didn't see this function ^^

data_processing/meteo/previsions_numeriques_temps/task_functions.py
estellebertrand: +1

utils/s3.py
estellebertrand: I created it because s3_client_kwargs variables as dict are used in several places in the repo : since in airflow we shouldn't instanciate a class - especially a costly one like an s3 client - during DAG parse time but only runtime, so this dict is passed to task later (meaning instanciation happens sometimes several times during a DAG lifetime, maybe can be improved but not the goal of this PR).

But because this was just a dict, Pylance would throw typing errors about the positional args. That's why I made it a typed dict so when writing s3_client_kwargs as a dict no mistake can be made and typing works well. Does it make sense or this is still too overkill for you ?

verticales/simplifions/tests/conftest.py
estellebertrand: Yes indeed i forgot that part ^^' I'll reach out to know more

estellebertrand:
> Thank you!!
> 
> Old naming `S3_URL_*` was not a good one, it was not a URL but a part of an URL which was a bit misleading and error-prone, good think we're getting rid of it!
> 
> A general remark: PR description says prod connexions are S3_RBX, S3_SBG, S3_EU_WEST_PAR, but the code uses S3_OVH_RBX, S3_OVH_SBG, S3_OVH_EU_WEST_PAR. Which one is right?

S3_OVH_* are the good ones, I forgot to update the PR description ^^ good catch thanks !

utils/s3.py
bolinocroustibat: Makes total sense! Let's keep it

bolinocroustibat (approved):
LGTM thank you! Just wondering about the tests [here](https://github.com/datagouv/datagouvfr_data_pipelines/pull/725/changes#r3683967046), maybe we could use Airflow connection mocks in tests.

utils/s3.py
estellebertrand: I'd say for now we're fine because we only use user who have enough permissions for production ! But might be a future improvement, unsure what would be the consequences though (for any other method applied after)

verticales/simplifions/tests/conftest.py
estellebertrand: So here DAG owners tell it's ok to break those tests for now, we can let as such. If it raises an error we can fix it afterwards but not critical !

estellebertrand:
Thank you @bolinocroustibat for your review :)

PR #726, opened 2026-07-31
add dag culture deps

eudespeyre:
Ajout du traitement DEPS dans la verticale culture

### DIFF
--- a/verticales/culture/deps/dag.py
+++ b/verticales/culture/deps/dag.py
@@ -9,7 +9,7 @@ from datagouvfr_data_pipelines.verticales.culture.deps.task_functions import (
  
 with DAG(
     dag_id="verticale_culture_deps",
-    # tous les jours à 2h, décalé d'une heure par rapport à `verticale_culture`
+    # very day at 2 a.m., offset by one hour from `verticale_culture`
     schedule="0 2 * * *",
     start_date=datetime(2026, 8, 1),
     catchup=False,
--- a/verticales/culture/deps/dag.py
+++ b/verticales/culture/deps/dag.py
@@ -1,0 +1,20 @@
+from datetime import datetime, timedelta
+ 
+from airflow.sdk import DAG
+from datagouvfr_data_pipelines.verticales.culture.deps.task_functions import (
+    get_perimeter,
+    notification,
+    refresh_tops,
+)
+ 
+with DAG(
+    dag_id="verticale_culture_deps",
+    # tous les jours à 2h, décalé d'une heure par rapport à `verticale_culture`
+    schedule="0 2 * * *",
+    start_date=datetime(2026, 8, 1),
+    catchup=False,
+    dagrun_timeout=timedelta(minutes=30),
+    tags=["verticale", "culture", "deps"],
+):
+    get_perimeter() >> refresh_tops() >> notification()
+ 
--- a/verticales/culture/deps/task_functions.py
+++ b/verticales/culture/deps/task_functions.py
@@ -1,6 +1,6 @@
 import logging
 
-import requests
+import pandas as pd
 from airflow.sdk import task
 from datagouvfr_data_pipelines.utils.datagouv import local_client
 from datagouvfr_data_pipelines.utils.grist import GristTable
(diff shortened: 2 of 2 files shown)


verticales/culture/deps/dag.py
```diff
@@ -0,0 +1,20 @@
...
+from airflow.sdk import DAG
+from datagouvfr_data_pipelines.verticales.culture.deps.task_functions import (
+    get_perimeter,
+    notification,
+    refresh_tops,
+)
+ 
+with DAG(
+    dag_id="verticale_culture_deps",
+    # tous les jours à 2h, décalé d'une heure par rapport à `verticale_culture`
```
maudetes, line 12: I think we use english for comments inline

verticales/culture/deps/task_functions.py
```diff
@@ -0,0 +1,95 @@
...
+from datagouvfr_data_pipelines.utils.datagouv import (
+    DATAGOUV_SECRET_API_KEY,
+    local_client,
+)
+from datagouvfr_data_pipelines.utils.grist import GristTable
+from datagouvfr_data_pipelines.utils.tchap import send_message
+
+# topic "univers-culture-deps" : source du périmètre, curé dans data.gouv.fr.
+# Le DAG le lit, il ne l'écrit jamais.
+topic_id = "69f2fca0f4f30af95d4bab8a"
```
maudetes, line 14: It doesn't exist on demo, right? Should it?

verticales/culture/deps/task_functions.py
```diff
@@ -0,0 +1,95 @@
...
+
+@task()
+def get_perimeter(**context):
+    """Identifiants des jeux de données rattachés au topic. Le contenu d'un topic
+    est exposé par la sous-ressource paginée `/elements/` ; un élément peut
+    pointer vers une URL externe ou être marqué « donnée non trouvée », d'où le
+    `element` potentiellement nul."""
+    ids = []
+    url = f"{local_client.base_url}/api/2/topics/{topic_id}/elements/?page_size=100"
+    while url:
```
maudetes, line 42: Maybe you could use a datagouv client that has a pagination helper. See [this snippet](https://github.com/datagouv/datagouvfr_data_pipelines/blob/2b27962bb4e2ce17924695d38b981880f826e29b/data_processing/meteo/pnt_monitor/task_functions.py#L65-L70) for example.

verticales/culture/deps/task_functions.py
```diff
@@ -0,0 +1,95 @@
...
+        url = payload.get("next_page")
+
+    if not ids:
+        raise ValueError("Aucun jeu de données dans le topic univers-culture-deps")
+    logging.info(f"> {len(ids)} jeux de données dans le périmètre DEPS")
+    context["ti"].xcom_push(key="datasets", value=sorted(set(ids)))
+
+
+@task()
+def refresh_tops(**context):
```
maudetes, line 59: Don't we want to mutualize the logic with [the one here](https://github.com/datagouv/datagouvfr_data_pipelines/blob/2b27962bb4e2ce17924695d38b981880f826e29b/verticales/culture/task_functions.py#L165)?
The approach seems different when the goal is similar right?

maudetes (dismissed):
Thank you for this PR! Seems to me that it is close to the main culture DAG but slightly different.
Maybe, it would be more maintainable to merge both?

I'll let @estellebertrand take a look at this as well

verticales/culture/deps/task_functions.py
eudespeyre: No, it doesn't exist on demo. There is currently no demo equivalent planned, as this topic is curated directly on data.gouv.fr and used as the source of truth for the DEPS scope. The DAG only reads it and never writes to it.

eudespeyre:
Hello @maudetes 
Thank you,

Yes, you're right. The only difference here should be how the perimeter is built. Once we have the dataset IDs from the DEPS topic, we can reuse the same catalog-based approach as verticale_culture. I'll align the implementation.

verticales/culture/deps/dag.py
```diff
@@ -0,0 +1,20 @@
...
+from airflow.sdk import DAG
+from datagouvfr_data_pipelines.verticales.culture.deps.task_functions import (
+    get_perimeter,
+    notification,
+    refresh_tops,
+)
+ 
+with DAG(
+    dag_id="verticale_culture_deps",
+    # very day at 2 a.m., offset by one hour from `verticale_culture`
```
estellebertrand, line 12:
```suggestion
    # every day at 2 a.m., offset by one hour from `verticale_culture`
```

typo

estellebertrand (dismissed):
The logic looks good - before merging, @eudespeyre can you confirm this has been tested locally with success ? I didn't run it on my side ;)

estellebertrand:
> Thank you for this PR! Seems to me that it is close to the main culture DAG but slightly different. Maybe, it would be more maintainable to merge both?
> 
> I'll let @estellebertrand take a look at this as well

I never read the other DAG - i'll check

estellebertrand (commented):
Before merging, agree with @maudetes  , what's the difference with the task `refresh_datasets_tops` ?

eudespeyre:
I would suggest refactoring the latter to rely on data.gouv.fr Topics, as introduced in [#726](https://github.com/datagouv/datagouvfr_data_pipelines/pull/726). The corresponding code is currently being prepared. This would make the approach more generic and allow us to reuse the same process for other thematic universes in the future.

estellebertrand:
> I would suggest refactoring the latter to rely on data.gouv.fr Topics, as introduced in [#726](https://github.com/datagouv/datagouvfr_data_pipelines/pull/726). The corresponding code is currently being prepared. This would make the approach more generic and allow us to reuse the same process for other thematic universes in the future.

If I understand well, you suggest to refactor the existing DAG so it uses topics instead of organisation for its selection of dataset tops ?

Modifying lines 182 to 184 in ``verticales/culture/tasks_function.py`  : 

```python
  datasets_catalog = datasets_catalog.loc[
        datasets_catalog["organization_id"].isin(orgas)
    ]
```
to rather rely on the list of dataset ids returned by your function `get_perimeter()` in this current PR ? or even replace the whole final task `refresh_datasets_tops` with the one from your PR? 

Also, I'm not familiar with this Grist table, does it serve as a data source for a prod website ?

PR #741, opened 2026-09-09
fix(dvf): compute stats for Paris, Lyon and Marseille

ThibaudDauce:
- Part of https://github.com/datagouv/data.gouv.fr/issues/2087
- See also https://github.com/datagouv/explore.data.gouv.fr/pull/223/ and https://github.com/datagouv/api-dvf/pull/8

DVF mutations carry the code of their municipal arrondissement (751xx, 693xx, 132xx) and never the code of the commune itself, so Paris, Lyon and Marseille were the only communes in France without any statistics at the commune scale.

Each mutation of an arrondissement is now also attached to its commune, which keeps medians computed on the mutations themselves rather than aggregated from the arrondissement medians. The other scale codes are cleared on those duplicates so they are still counted once per departement, EPCI and section, and national stats keep using the untouched mutations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,10 @@ jobs:
             python -m venv venv
             . venv/bin/activate
             pip install --upgrade pip
-            pip install ruff>=0.11.2
+            # version alignée sur .pre-commit-config.yaml : sans borne haute, une
+            # nouvelle version de ruff élargit les règles par défaut et met la CI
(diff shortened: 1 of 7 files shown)


estellebertrand (commented):
LGTM, but have you tested it locally ?

ThibaudDauce:
> LGTM, but have you tested it locally ?

Yes I tested locally and it works

PR #743, opened 2026-09-11
chore(dashboard): remove old support URL

maudetes:
Site id 176 in matomo was about https://support.data.gouv.fr/ that doesn't exist anymore


estellebertrand (approved):
Maybe just linting / ruff ?

PR #745, opened 2026-09-15
feat(meta): add a notification exclusion to the meta dag

hacherix:
* Remove obsolete SIRENE DAGs from the list and add `data_processing_dfi_consolidation`
* Sort the DAGs in the Tchap notification alphabetically with the failed ones always last
* Add an exclusion feature so we can override, add new DAG owner and remove default DAG owner from the notifications


estellebertrand (approved):
LGTM ;)

